PR #6, opened 2021-02-10
Cardigann auth

Qstick:
#### Database Migration
NO

#### Description
Port over ugly Jackett Login code for Cardigann Auth support

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Common/Http/HttpResponse.cs
+++ b/src/NzbDrone.Common/Http/HttpResponse.cs
@@ -11,18 +11,20 @@ namespace NzbDrone.Common.Http
     {
         private static readonly Regex RegexSetCookie = new Regex("^(.*?)=(.*?)(?:;|$)", RegexOptions.Compiled);
 
-        public HttpResponse(HttpRequest request, HttpHeader headers, byte[] binaryData, HttpStatusCode statusCode = HttpStatusCode.OK)
+        public HttpResponse(HttpRequest request, HttpHeader headers, CookieCollection cookies, byte[] binaryData, HttpStatusCode statusCode = HttpStatusCode.OK)
         {
             Request = request;
             Headers = headers;
+            Cookies = cookies;
             ResponseData = binaryData;
             StatusCode = statusCode;
         }
 
-        public HttpResponse(HttpRequest request, HttpHeader headers, string content, HttpStatusCode statusCode = HttpStatusCode.OK)
+        public HttpResponse(HttpRequest request, HttpHeader headers, CookieCollection cookies, string content, HttpStatusCode statusCode = HttpStatusCode.OK)
         {
             Request = request;
             Headers = headers;
+            Cookies = cookies;
             ResponseData = Headers.GetEncodingFromContentType().GetBytes(content);
             _content = content;
             StatusCode = statusCode;
@@ -30,6 +32,7 @@ namespace NzbDrone.Common.Http
 
         public HttpRequest Request { get; private set; }
         public HttpHeader Headers { get; private set; }
+        public CookieCollection Cookies { get; private set; }
         public HttpStatusCode StatusCode { get; private set; }
         public byte[] ResponseData { get; private set; }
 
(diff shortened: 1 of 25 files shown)


src/NzbDrone.Common/Http/HttpResponse.cs
```diff
@@ -11,25 +11,28 @@ public class HttpResponse
...
             ResponseData = binaryData;
             StatusCode = statusCode;
         }
 
-        public HttpResponse(HttpRequest request, HttpHeader headers, string content, HttpStatusCode statusCode = HttpStatusCode.OK)
+        public HttpResponse(HttpRequest request, HttpHeader headers, CookieCollection cookies, string content, HttpStatusCode statusCode = HttpStatusCode.OK)
         {
             Request = request;
             Headers = headers;
+            Cookies = cookies;
```
Qstick, line 27: Added CookieCollection to HttpResponse due to issue with arr Http Stack and getting cookies from Headers. Since they are stored in a dictionary in HttpResponse only the first cookie is ever available due to the limitation that there can only be one header key of a given value. Thus if the HttpWebRequest returns 3 cookies all of which will have header key of `Set-Cookie` only one is loaded in the dictionary. This is no good when we depend on storing cookies for Auth

src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannParser.cs
```diff
@@ -333,8 +333,8 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                                     continue;
                                 }
 
-                                _logger.Error("Error while parsing field={0}, selector={1}, value={2}: {3}", field.Key, field.Value.Selector, value == null ? "<null>" : value, ex.Message);
-                                throw;
+                                //Parser errors usually happen on every result and are costly to performance, trace only
+                                _logger.Trace("Error while parsing field={0}, selector={1}, value={2}: {3}", field.Key, field.Value.Selector, value == null ? "<null>" : value, ex.Message);
```
Qstick, line 337: This make parsing extremely slow if any of the rules don't parse. Ex. Blutopia and posters

src/NzbDrone.Core/Indexers/Definitions/PassThePopcorn/PassThePopcornParser.cs
```diff
@@ -49,6 +49,8 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                 throw new IndexerException(indexerResponse, $"Unexpected response header {indexerResponse.HttpResponse.Headers.ContentType} from API request, expected {HttpAccept.Json.Value}");
             }
 
+            CookiesUpdater(indexerResponse.HttpResponse.GetCookies(), DateTime.Now + TimeSpan.FromDays(30));
```
Qstick, line 52: Remove, was only for test

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -373,7 +370,25 @@ protected virtual ValidationFailure TestConnection()
                 {
                     _indexerStatusService.UpdateCookies(Definition.Id, cookies, expiration);
                 };
+
                 var generator = GetRequestGenerator();
+                generator.GetCookies = () =>
```
Qstick, line 375: Use function above

src/NzbDrone.Core/Indexers/IndexerStatusService.cs
```diff
@@ -54,12 +54,15 @@ public void UpdateRssSyncStatus(int indexerId, ReleaseInfo releaseInfo)
 
         public void UpdateCookies(int indexerId, IDictionary<string, string> cookies, DateTime? expiration)
         {
-            lock (_syncRoot)
+            if (indexerId > 0)
```
Qstick, line 57: This will prevent indexers with 0 from being added to table during test call before add

PR #65, opened 2021-05-21
Add animetosho

bakerboy448:
#### Database Migration
NO

#### Description
Untested; theory should work

Ref
https://animetosho.org/about

Newznab/Torznab links to: `https://feed.animetosho.org/api`
>Is there a feed available?
>RSS2 (torrents only), Newznab/Torznab and ATOM. If you are trying to add any of these to an application which insists on an API key, please just put 0 as the key (or anything else, as Anime Tosho doesn't require an API key, but it's helpful if everyone uses the same key for server caching).
Is there an API available?
>The feed (see above) can serve as an API. You can also replace the rss2 or atom in the feed URL with json to get a JSON formatted version. The JSON API also supports querying torrent/file info if given a supplied torrent ID (example).
>Alternatively, daily database exports are also available for download here.

#### Todos
- Tests-N/A
- Translation Keys-N/A
- Wiki Updates-N/A

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Newznab/Newznab.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Newznab/Newznab.cs
@@ -84,6 +84,7 @@ namespace NzbDrone.Core.Indexers.Newznab
             {
                 yield return GetDefinition("abNZB", GetSettings("https://abnzb.com"));
                 yield return GetDefinition("altHUB", GetSettings("https://althub.co.za"));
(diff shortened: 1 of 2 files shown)


Qstick:
The /api is the default API sub and thus gets added automatically. You don't need to include it.

bakerboy448:
fixed up

Qstick:
I think these being named the same will cause front end list issues due to having duplicate keys

Qstick (approved):

PR #155, opened 2021-06-08
Add detailed help text for sync levels and some wording stuff

austinwbest:
Adjust donation buttons to fit on mobile
Other translation updates


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -352,6 +356,9 @@
   "Style": "Style",
   "SuggestTranslationChange": "Suggest translation change",
   "SyncAppIndexers": "Sync App Indexers",
+  "SyncLevel": "Sync Level",
+  "SyncLevelAddRemove": "Add and Remove Only: When it is added or removed from Prowlarr, it will update this remote application.",
+  "SyncLevelFull": "Full Sync: When anything changes in this remote application, Prowlarr will keep it in sync. An example would be to delete from this remote application means Prowlarr will put it back.",
```
bakerboy448, line 361: Would it be too much to squeeze in that it impacts categories too?

austinwbest: It is merely meant to be an example to trigger people to think that it will keep it in sync. It is quite long enough as it is but if someone feels the need to tack on more later lol, feel free!

RobinDadswell: "Prowlarr will keep it in sync" this to me sounds like it would take the config back to prowlarr (which it doesn't obviously) but not really sure how it could be changed so I would take that this is enough already

austinwbest: Yea, wording can be changed by anyone who feels it could be relayed in a clearer way (wont bother me a bit as i know what it does lol). Just wanted to get something in place so people stop asking.

Qstick: "When anything changes in this remote application, Prowlarr will keep it in sync." Agree this is a bit confusing and doesn't really catch the full scope of full sync (changes made in Prowlarr that push out)

How about something to this effect:

Full sync will keep your app and Prowlarr fully in sync. Any change made in Prowlarr is then synced to the other program. Any change made remotely will be overridden by Prowlarr on next sync

austinwbest: image

I tried to trim it up a little so it wasn't so lengthy in the UI. Everybody happy with that?

PR #178, opened 2021-06-09
Implements the ability to filter Indexers by Language & Privacy

Lagicrus:
#### Database Migration
NO

#### Description
Adds the ability to filter the "Add Indexers" modal by the Language and the Privacy setting of an indexer.

https://user-images.githubusercontent.com/24227350/121312806-a6768b80-c8fd-11eb-9f2d-02118db83706.mp4

#### Todos
- [x] Translation Keys


Lagicrus:
There are two main issues atm, one more easily solvable than the other
- Styling, when you sort, the widths change
- Data with Redux

Currently, this is fine operating within itself, but when you add the sorting with Redux in the Table Header, it resets filtering.
It does feel a bit odd to control something as basic as this with Redux, when it could just work the same way this all does, which I am tempted to implement, but unsure if that is further out of scope.

I am open for ideas and feedback

Qstick:
This looks pretty neat. I agree with the styling bit on the widths, if we could fix that it would be great.

RE the filter/sort issue, Can you hook filtering into redux similar to how we do it elsewhere?  The plan eventually is to virtualize this table to pick up a bit of render speed, similar to the main table and search table.

Would we also want to add this to the torrent/usenet column?

Lagicrus:
> This looks pretty neat. I agree with the styling bit on the widths, if we could fix that it would be great.
> 
> RE the filter/sort issue, Can you hook filtering into redux similar to how we do it elsewhere? The plan eventually is to virtualize this table to pick up a bit of render speed, similar to the main table and search table.
> 
> Would we also want to add this to the torrent/usenet column?

Styling has been fixed, adding in filtering by protocol. Unsure if it should be using "usenet" as the label or "nzb" (see below)

https://user-images.githubusercontent.com/24227350/123590997-62cbbf00-d7e3-11eb-9b64-b8ce142ad62f.mp4

Looking into Redux now & seeing if I can't fix a bug when you filter by two (or more) at once

Lagicrus:
Styling has been fixed, again, as it needed moving up to the headers as whilst it worked fine, the moment your filter returns 0 results (e.g. `ar-ar` `public`) the styling breaks. This reason was also the reason why I removed the check for length.
Reworked the internal logic to allow for multiple filters to be applied at once as shown below.

https://user-images.githubusercontent.com/24227350/123597690-8eeb3e00-d7eb-11eb-81c1-f407f3624e60.mp4

I'll begin looking at implementing this in Redux. Though I do have one question @Qstick , what should it do if there are 0 results? 
- Custom "row" showing no results?
- Message outside of the table?
- Image?

Qstick:
We use message outside of table in other places, So let's do that.

Lagicrus:
I think this is ready for review now 👌 
Added an info message for when there are no results - added into translate files as well
Sorting works now as well
Open to feedback

bakerboy448:
https://github.com/Prowlarr/Prowlarr/blob/develop/CONTRIBUTING.md#contributing-code
> **Rebase from Prowlarr's develop branch, don't merge**

Lagicrus:
> https://github.com/Prowlarr/Prowlarr/blob/develop/CONTRIBUTING.md#contributing-code
> 
> > **Rebase from Prowlarr's develop branch, don't merge**

🤦‍♂️ my bad, just wanted to resync up the two. Would a revert work best on that commit?

bakerboy448:
needs a quick rebase with en.json

jzeirm:
Is it possible this might also include a filter to hide indexers already added?

Lagicrus:
Well, I told it to rebase, and this happened.. Git is fun
Shall I leave it like this or adjust it somehow?

csandman:
I just made my own PR for this feature today because I didn't realize this existed, obviously a lot more work went into this one than mine haha.

One thing I prefer about mine, however, is that the filters use the `<EnhancedSelectInput />` component so you can select multiple options in each category at once.  This is specifically useful for the languages field, but It seems like it would be fine to use it for all of them.  An example of how someone might use this, is viewing all public `en-` language indexers.

Besides that, it would be great if this feature could get merged soon, this is something I'd definitely use regularly.

bakerboy448:
closing in lieu of https://github.com/Prowlarr/Prowlarr/commit/f69f96695b1262a89ff1ea4ff2de376a55925b99 / #607

PR #211, opened 2021-06-14
New: Better Application Tests

### LINKED ISSUE #209: Run Indexer test during App add to test App to Prowlarr Connection
**Is your feature request related to a problem? Please describe.**

Users seem to have issue with app configuration as a result of App not being able to connect pack to Prowlarr. Currently we only test if Prowlarr can connect to app during testing a new App.

**Describe the solution you'd like**

Hit Arr indexer test endpoint with fake data on add to test if App can connect back, don’t allow the add if it can’t.

Qstick:
#### Database Migration
NO

#### Description
Instead of just calling `System/Status` on the application to test connection, this extends the tests to do a mock indexer test call. This ensures the application calls back to Prowlarr to test the indexer connection thus testing both the connection to the app and the connection back to Prowlarr. This should eliminate or minimize the reports sync failing due to bad application settings.

This also adds code to the Newznab endpoint to return mock results if calling with indexer id of 0

* Fixes #209

### DIFF
--- a/src/NzbDrone.Core/Applications/Lidarr/Lidarr.cs
+++ b/src/NzbDrone.Core/Applications/Lidarr/Lidarr.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using FluentValidation.Results;
 using Newtonsoft.Json.Linq;
 using NLog;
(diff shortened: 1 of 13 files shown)


RobinDadswell (commented):
Looks alright to me, although I think it is probably worth adding to health checks as well to make situations where things are different easier to troubleshoot (although that would duplicate health check warnings between *Arr/Prowlarr)

PR #240, opened 2021-06-18
(Indexers) Add Redacted API Support

### LINKED ISSUE #76: (Indexer) - Torrent Redacted - APIKey Support
title, for tracking

[AB#31](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/31)

ntldr0:
#### Database Migration
YES

#### Description
Switch Redacted implementation over to API-based.  Using Jackett, for example, will lead to IP bans.  However, Lidarr's API implementation is stable and doesn't result in IP bans.

Implementation is based on [Lidarr](https://github.com/lidarr/Lidarr/tree/3523eb63945acbe16a78b320690ea8b3b333fa84/src/NzbDrone.Core/Indexers/Redacted).

#### Todos
- [ ] Tests
- [ ] Translation Keys
- [ ] Wiki Updates

* Fixes #76


ntldr0:
@Qstick I think this PR is ready for review.  The changes seem to be working with both Lidarr and Readarr.

Qstick:
Thanks I'll take a look when I can

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
```diff
@@ -41,13 +74,286 @@ protected override IndexerCapabilities SetCapabilities()
...
+            Apikey = "";
+            Passkey = "";
+            UseFreeleechToken = false;
+        }
+
+        [FieldDefinition(1, Label = "API Key", HelpText = "Redacted API Key")]
+        public string Apikey { get; set; }
+
+        [FieldDefinition(1, Hidden = HiddenType.Hidden)]
+        public string Passkey { get; set; }
```
Qstick, line 349: We shouldn't need a field definition for this, it's never seen or used as a field.

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
```diff
@@ -1,23 +1,56 @@
...
+using NzbDrone.Core.IndexerSearch.Definitions;
 using NzbDrone.Core.Messaging.Events;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.ThingiProvider;
+using NzbDrone.Core.Validation;
 
 namespace NzbDrone.Core.Indexers.Definitions
 {
-    public class Redacted : Gazelle.Gazelle
+    public class Redacted : TorrentIndexerBase<RedactedSettings>
```
Qstick, line 22: Changing the ConfigContract breaks existing instances, this needs a migration to avoid cast errors

image

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
```diff
@@ -1,23 +1,56 @@
...
+        {
+            return new RedactedRequestGenerator() { Settings = Settings, Capabilities = Capabilities, BaseUrl = BaseUrl, HttpClient = _httpClient };
+        }
+
+        public override IParseIndexerResponse GetParser()
+        {
+            return new RedactedParser(Settings, Capabilities.Categories, BaseUrl);
+        }
+
+        protected override bool CheckIfLoginNeeded(HttpResponse httpResponse)
```
Qstick, line 48: No use in this override if id just returns `false` that is default behavior.

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
```diff
@@ -1,23 +1,56 @@
...
+    public class Redacted : TorrentIndexerBase<RedactedSettings>
     {
         public override string Name => "Redacted";
         public override string BaseUrl => "https://redacted.ch/";
+        public override string Description => "Redacted is a Private Torrent Tracker for Music";
+        public override string Language => "en-us";
+        public override Encoding Encoding => Encoding.UTF8;
+        public override DownloadProtocol Protocol => DownloadProtocol.Torrent;
         public override IndexerPrivacy Privacy => IndexerPrivacy.Private;
+        public override IndexerCapabilities Capabilities => SetCapabilities();
```
Qstick, line 31: We can also set SupportsRedirect to true for this given it builds standalone download links that don't require cookies to work

`public override bool SupportsRedirect => true;`

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
```diff
@@ -41,13 +74,286 @@ protected override IndexerCapabilities SetCapabilities()
...
+
+            pageableRequests.Add(GetRequest(searchCriteria.SanitizedSearchTerm));
+
+            return pageableRequests;
+        }
+
+        private IEnumerable<IndexerRequest> GetRequest(string searchParameters)
+        {
+            // Retrieve passkey if not already done
+            if (string.IsNullOrWhiteSpace(Settings.Passkey))
```
Qstick, line 140: Does this passkey ever expire? if so there is no way for it to be re-generated the way the code is written.

Qstick (changes requested):

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
ntldr0: Since this is the same passkey used in the torrent files themselves, no, this should never expire.  However, a user can manually go into their RED profile and manually reset the passkey.  But, that actually shouldn't cause any problems here.  As far as I can tell, the passkey is never committed to the database, and instead, persists only for this particular instance of  RedactedSettings object.  In practice, this seems to fetch the passkey once per query.  But the passkey shouldn't get fetched for each page of results, only once.

This definitely feels a bit hacky though, so I am very open to suggestions on better ways to manage the passkey.

Qstick: Yea, I'd prefer to keep it automated, the less we rely on the user the better

bakerboy448:
conflicts; needs a rebase

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
Junkbite: i dont see how this is any different than having to end the api key for united based trackers like blutopia and even beyondhd. beyondhd you have to type in the api key and rss key. id rather have to add the api key like any other tracker. its been this way for a long time why change to this automated way that may be more of a headache to get working again if a user deletes their api key or regenerates it?

ntldr0:
@bakerboy448 rebased.

src/NzbDrone.Core/Indexers/Definitions/Redacted.cs
ntldr0: I removed the `if (string.IsNullOrWhiteSpace(Settings.Passkey))` check, this way it is both automated and will continue to work even if the user generates a new passkey.

PR #311, opened 2021-07-02
If no categories are passed in, flag up a unknown error

Lagicrus:
#### Database Migration
NO

#### Description
Builds off previous work in #301 , but instead of just displaying nothing, it instead shows a placeholder category that has a tooltip explaining to the user that they should report it, so it can get fixed.

image
image

#### Todos
- [ ] Translation Keys

### DIFF
--- a/frontend/src/Search/Table/CategoryLabel.js
+++ b/frontend/src/Search/Table/CategoryLabel.js
@@ -7,7 +7,7 @@ import Tooltip from '../../Components/Tooltip/Tooltip';
 function CategoryLabel({ categories }) {
   const sortedCategories = categories.filter((cat) => cat.name !== undefined).sort((c) => c.id);
 
-  if (categories.length === 0) {
+  if (categories?.length === 0) {
     return (
       <Tooltip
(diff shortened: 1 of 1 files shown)


frontend/src/Search/Table/CategoryLabel.js
```diff
@@ -1,10 +1,22 @@
...
 import React from 'react';
 import Label from 'Components/Label';
+import { kinds, tooltipPositions } from 'Helpers/Props';
+import Tooltip from '../../Components/Tooltip/Tooltip';
 
-function CategoryLabel({ categories = [] }) {
+function CategoryLabel({ categories }) {
   const sortedCategories = categories.filter((cat) => cat.name !== undefined).sort((c) => c.id);
 
+  if (categories.length === 0) {
```
Qstick, line 10: This will error if categories comes in null/undefined right? I think that was the reason for the default prop

Lagicrus: Fair point, tested your concerns and it was line 8 that broke first due to filter, so I added back in `defaultProps` to resolve that 👌

Qstick (approved):

PR #319, opened 2021-07-05
Reworked the Search Page for mobile to improve speed & usability

Lagicrus:
#### Database Migration
NO

#### Description
On the mount of the search page, untick "non-essential" columns, leaving the user with the title and which indexer.
During testing of performance, we go from a previous average of 30ms~ per VirtualTableRow to 16ms~ when using the Mid-Tier Mobile preset which slows my CPU by 4x

Example of the changed options
image
Example of the search results using Chrome's Pixel 2 XL preset
image

### DIFF
--- a/frontend/src/Search/SearchIndex.js
+++ b/frontend/src/Search/SearchIndex.js
@@ -49,6 +49,16 @@ class SearchIndex extends Component {
   }
 
   componentDidMount() {
+    const { isSmallScreen, columns, onTableOptionChange } = this.props;
+    if (isSmallScreen) {
+      const mobileColumns = _.cloneDeep(columns);
+
+      for (const columnName of ['protocol', 'age', 'size', 'grabs', 'peers', 'category', 'indexerFlags']) {
+        const column = _.find(mobileColumns, { name: columnName });
+        column.isVisible = false;
+      }
+      onTableOptionChange({ columns: mobileColumns });
+    }
     this.setJumpBarItems();
 
     window.addEventListener('keyup', this.onKeyUp);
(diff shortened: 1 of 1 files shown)


frontend/src/Search/SearchIndex.js
```diff
@@ -49,6 +49,16 @@ class SearchIndex extends Component {
...
   componentDidMount() {
+    const { isSmallScreen, columns, onTableOptionChange } = this.props;
+    if (isSmallScreen) {
+      const mobileColumns = _.cloneDeep(columns);
+
+      for (const columnName of ['protocol', 'age', 'size', 'grabs', 'peers', 'category', 'indexerFlags']) {
+        const column = _.find(mobileColumns, { name: columnName });
+        column.isVisible = false;
+      }
+      onTableOptionChange({ columns: mobileColumns });
```
Qstick, line 60: Doesn't this kill all customization on mobile? Seems like it would never let you turn these columns on even if you wanted to. As it's always going to reset them on a page load

Lagicrus: This is true, but unsure how to fix it.
The only solution that comes to mind is prompting the user to "change their settings for mobile" or  "stopping the automated optimisation".
Neither of which is great and unsure this second if we have a component that works for this or if one would need creating.

There is also the concept of having a bool flag that signals that a user has overridden the config and to use that instead. But that would require each device to have its own configuration and I am unsure if that's the case?

But I think the core issue here is that "saving the customisation" doesn't seem to be a thing in live?
On PC I loaded up the search page, opened settings, removed protocol. Then reloaded the page. And it was checked again. This makes me think this feature isn't in or it is bugged.

PR #339, opened 2021-07-11
Fixed: Respect categories for ImmortalSeed in search

PearsonFlyer:
#### Database Migration
NO

#### Description
ImmortalSeed was returning results from all categories, but now respects category selection in Prowlarr searches.

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/ImmortalSeed.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/ImmortalSeed.cs
@@ -171,10 +171,14 @@ namespace NzbDrone.Core.Indexers.Definitions
         {
             var searchUrl = Settings.BaseUrl + "browse.php";
 
-            //TODO - Actually map some categories here
             if (term.IsNotNullOrWhiteSpace())
             {
                 searchUrl += string.Format("?do=search&keywords={0}&search_type=t_name&category=0&include_dead_torrents=no", WebUtility.UrlEncode(term));
+            }
+
+            if (categories != null && categories.Length > 0)
+            {
+                searchUrl += "&selectedcats2=" + string.Join(",", Capabilities.Categories.MapTorznabCapsToTrackers(categories));
             }
 
             var request = new IndexerRequest(searchUrl, HttpAccept.Html);


src/NzbDrone.Core/Indexers/Definitions/ImmortalSeed.cs
```diff
@@ -171,12 +171,20 @@ private IEnumerable<IndexerRequest> GetPagedRequests(string term, int[] categori
         {
             var searchUrl = Settings.BaseUrl + "browse.php";
 
-            //TODO - Actually map some categories here
             if (term.IsNotNullOrWhiteSpace())
             {
                 searchUrl += string.Format("?do=search&keywords={0}&search_type=t_name&category=0&include_dead_torrents=no", WebUtility.UrlEncode(term));
```
Qstick, line 176: We should remove category=0 here if it's not needed now

src/NzbDrone.Core/Indexers/Definitions/ImmortalSeed.cs
```diff
@@ -171,12 +171,20 @@ private IEnumerable<IndexerRequest> GetPagedRequests(string term, int[] categori
...
             }
 
+            searchUrl += "&selectedcats2=";
+
+            foreach (var cat in Capabilities.Categories.MapTorznabCapsToTrackers(categories))
+            {
+               searchUrl += cat + "%2C";
+            }
+
+            searchUrl = searchUrl.Substring(0, searchUrl.Length - 3);
```
Qstick, line 186: We could probably cleanup this block a bit using a join to avoid having to cleanup afterwards

Qstick (changes requested):

src/NzbDrone.Core/Indexers/Definitions/ImmortalSeed.cs
PearsonFlyer: Weirdly, "&category=0" is still needed, but it's not actually categories.

Qstick: Ah... Odd

src/NzbDrone.Core/Indexers/Definitions/ImmortalSeed.cs
Qstick: Try this

```
if (categories != null && categories.Length > 0)
{
    searchUrl += "&selectedcats2=" + string.Join(",", Capabilities.Categories.MapTorznabCapsToTrackers(categories));
}
```

Qstick (approved):

PR #351, opened 2021-07-14
Fixed (Anime Tosho): Only include Releases which have a NZB

### LINKED ISSUE #200: (Indexer) - AnimeTosho (Usenet) - Torrent links w/o NZBs in Usenet results
Hi, it seems prowlarr gets confused if the indexer one or more links to download or if the link does not have nzb download link, in this case, animetosho is torrent/nzb indexer, sometimes links do not have nzb download links. in that case i think the results should be rejected.  i have attached trace log for when the issue happens

[prowlarr.trace.txt](https://github.com/Prowlarr/Prowlarr/files/6642140/prowlarr.trace.txt)

[AB#59](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/59)

DevYukine:
#### Database Migration
NO

#### Description
The Newsnab API of Anime Tosho also includes releases which have no NZB (either because its not processed yet or because it was handled before they started uploading to Usenet), i added filters which filter the releaes out if the Newsnab provider gives a type (which Anime Tosho does)

I tested it locally and it seems to not break any other Usenet indexer (tested with DrunkenSlug, NZBCat, SimplyNZBs)

#### Todos
- [ ] Tests

* Fixes #200

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Newznab/NewznabRssParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Newznab/NewznabRssParser.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Xml;
 using System.Xml.Linq;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Indexers.Exceptions;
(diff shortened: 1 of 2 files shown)


Qstick:
Should the same checks be in Torznab parser as well?

DevYukine:
I can make another PR with the same checks if you wish but i think its outside of the scope of this PR to add that to Torznab

Qstick (approved):

PR #376, opened 2021-07-31
Fixed: (Indexer) AnimeBytes improve sonarr compatibility and make optional

DevYukine:
#### Database Migration
NO

#### Description
This PR changes two things

1. If an Episode is found, ignore the Season matching since it confuses Sonarr too much and Episodes are going by Absolute Numbering anyway
2. Makes all the Season guessing optional with an by default enabled Checkbox, since its really only working in 70% of all cases and still has a lot of false positives when Anime have Seasonal names with no Season number in it (Jojo for example with  stardust crusaders)

Before:

`TSUKIPRO THE ANIMATION 2 S2 4`
`Tensei Shitara Slime Datta Ken 2nd Season Part 2 S2 40`

Now:

`TSUKIPRO THE ANIMATION 2 04`
`Tensei Shitara Slime Datta Ken 2nd Season Part 2 40`

Not needed

#### Todos
- [ ] Tests

None

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
@@ -252,82 +252,71 @@ namespace NzbDrone.Core.Indexers.Definitions
 
                     foreach (var torrent in group.Torrents)
                     {
-                        const string defaultReleaseInfo = "S01";
-                        var releaseInfo = defaultReleaseInfo;
-                        string episode = null;
+                        var releaseInfo = _settings.EnableSonarrCompatibility ? "S01" : "";
+                        int? episode = null;
                         int? season = null;
                         var editionTitle = torrent.EditionData.EditionTitle;
                         if (!string.IsNullOrWhiteSpace(editionTitle))
                         {
                             releaseInfo = WebUtility.HtmlDecode(editionTitle);
 
-                            var simpleSeasonRegEx = new Regex(@"Season (\d+)", RegexOptions.Compiled);
-                            var simpleSeasonRegExMatch = simpleSeasonRegEx.Match(releaseInfo);
-                            if (simpleSeasonRegExMatch.Success)
+                            if (_settings.EnableSonarrCompatibility)
                             {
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
```diff
@@ -514,6 +504,9 @@ public AnimeBytesSettings()
         [FieldDefinition(3, Label = "Username", HelpText = "Site Username", Privacy = PrivacyLevel.UserName)]
         public string Username { get; set; }
 
+        [FieldDefinition(4, Label = "Enable Sonarr Compatibility", Type = FieldType.Checkbox,  HelpText = "Makes Prowlarr try to add Season information into Release names, without this Sonarr can't match any Seasons, but it has a lot of false positives as well")]
+        public bool EnableSonarrCompatibility { get; set; }
+
         [FieldDefinition(4)]
```
Qstick, line 510: Remaining fields need renumbered

Qstick:
What’s the case for making this optional instead of just doing it all the time?

DevYukine:
The main reason is that the season matching and default S1 thing only works maybe like 70% of the time, and the other 30% it does wrongly assume its S1 and so Sonarr matches incorrect which results in wrongly downloading & importing and some people might just not want that, also its only a fix for sonarr, other software which prowlarr aims to support later on might not even want/need these since they don't totally rely on TVDB as Sonarr is.

So i think its better to move it behind an by default enabled (so its backward compatible) option 😄

bakerboy448:
you'll need a squash as well and need to cleanup the mess caused by merging instead of rebasing

always rebase, never merge ;)

DevYukine:
This was actually only a problem because develop was forced pushed and changed 3 commits which were in my history due to me merging develop back into this branch, i've now reset the history to be identical and cherry picked my commits back in.

Cleaned up the mess, remember kids never force push a develop branch 😉 

And Rebase would've had the same issue fyi

Qstick (approved):

PR #390, opened 2021-08-03
Fixed: (Indexer) Rutracker - multiple languages support

chryzsh:
#### Database Migration
NO

#### Description
We already add "rus" to all release names in Rutracker as all releases have Russian audio (see line 1635). However, a lot of releases also include the original language in either the format "Original xxx" where xxx is the language code, but also just "Original". This PR handles that latter scenario. By adding "multi" in the release name, *arr correctly identifies them as multi-language releases. For increased precision, the user can in *arr configure the "Multi Language" option on the indexer to "Original" (see screenshot) and get a correct language mapping for Rutracker releases. If the user does not select any option here, this PR has no effect other than indicating to *arr that this is a multi-language release, which is still correct.

image

#### Todos
- [ ] Tests

* No issue reported.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/RuTracker.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/RuTracker.cs
@@ -1635,6 +1635,12 @@ namespace NzbDrone.Core.Indexers.Definitions
                     release.Title += " rus";
                 }
 
+                // language fix: if releases contains Original track, add "multi"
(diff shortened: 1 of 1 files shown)


chryzsh:
Closing — this is 5 years old and upstream has since refactored the release-title processing into a dedicated `RuTrackerTitleParser` class, so the original patch no longer applies cleanly. Not worth reworking at this point.

PR #391, opened 2021-08-04
[Needs Work] New: (Indexer) Rarbg Alt Links

### LINKED ISSUE #333: Better support for RARBG indexer
**Is your feature request related to a problem? Please describe.**

- RARBG consistently returns no results for random periods of time. 
- There’s no retries currently, which should be done due to a very problematic API
- During arr RSS Sync, if there's zero results returned, it is **not** marked as a failure, when it should be
- IMDb/~~TVDB/TMDB~~ search doesn't work

Jackett, in comparison returns results somewhat better.

**Describe the solution you'd like**

- Support site links (pre-determined list of proxies/user-defined), retries like Jackett:

- Mark indexer as a failure if RSS Sync returns zero results and API returns `200`.
- Support IMDb ~~/TVDB/TMDB~~ search by auto detection. API docs:

81C1EFA8-12AF-42A3-A705-CDC010B8227A

**Describe alternatives you've considered**

**Additional context**

[AB#1248](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/1248)

bakerboy448:
#### Database Migration
NO

#### Description
Add Alt links
#### Todos
- n/a Tests
- n/a Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)
- n/a [Wiki Updates](https://wiki.servarr.com)
- [ ] fixup if it doesn't build; mobile commit

* Fixes #333

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Rarbg/Rarbg.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Rarbg/Rarbg.cs
@@ -16,7 +16,15 @@ namespace NzbDrone.Core.Indexers.Rarbg
         private readonly IRarbgTokenProvider _tokenProvider;
 
         public override string Name => "Rarbg";
-        public override string[] IndexerUrls => new string[] { "https://torrentapi.org" };
+        public override string[] IndexerUrls => new string[]
+        {
+          "https://torrentapi.org",
+          "https://rarbg.to",
+          "https://rarbgprx.org",
+          "https://proxyrarbg.org",
+          "https://rarbgget.org",
+          "https://rarbgproxied.org"
+        };
         public override string Description => "RARBG is a Public torrent site for MOVIES / TV / GENERAL";
 
         public override DownloadProtocol Protocol => DownloadProtocol.Torrent;


agneevX:
Hi Baker, would you consider adding these sites in Insecure HTTP 80 as well as I've found that, in my case, these load over port 80 but not over 443.

Screen Shot 2021-08-04 at 12 23 36 PM


Screen Shot 2021-08-04 at 12 23 26 PM

agneevX:
Also, I just realized that adding these URLs like this won't do anything since torrentapi is the only query API. 

I'm not exactly sure what [Jackett does with site URLs](https://github.com/Jackett/Jackett/blob/1f1f42cf597be8431396ed3333384c1a371b8349/src/Jackett.Common/Indexers/RarBG.cs) but it ends up quering torrentapi (over port 80 or 443) only.

src/NzbDrone.Core/Indexers/Definitions/Rarbg/Rarbg.cs
```diff
@@ -16,7 +16,15 @@ public class Rarbg : TorrentIndexerBase<RarbgSettings>
         private readonly IRarbgTokenProvider _tokenProvider;
 
         public override string Name => "Rarbg";
-        public override string[] IndexerUrls => new string[] { "https://torrentapi.org" };
+        public override string[] IndexerUrls => new string[] 
+        { 
+          "https://torrentapi.org",
+          "https://rarbg.to",
```
Qstick, line 22: I'm not sure that all of these actually work as API alternatives. They are more site/info url alternatives. The code here needs a few more adjustments to know when to use what

bakerboy448: Simple adjustments or something a bit more complex?

Qstick: Should be simple enough

bakerboy448: Q I'll leave the PR up, but defer to you or someone else where the specific changes are to be made and/or just pushing those up.

unfortunately don't have time this/next week for anything much more than quick easy commits

Qstick:
So we currently use the `info_page` property returned from torrentapi to direct to the info page

```
"info_page": "https://torrentapi.org/redirect_to_info.php
```

So I'm not really sure what we would do with alternative urls for rarbg unless they were alternatives for the api.

agneevX:
Yeah I guess this can be closed.

PR #402, opened 2021-08-06
[BugFix] (Tracker) Internet Archive: Add handling of missing result fields.

### LINKED ISSUE #401: [BUG]: Internet Archive: cannot save - Value cannot be null. (Parameter 'infoHash') - Sourced from btih

### Current Behavior

Testing the Internet Archive indexer that was recently added fails.

### Expected Behavior

I expect to be able to add the indexer to my instance without issues.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04
- Prowlarr: 0.1.0.761
- Docker Install: yes - nightly-632f276 by hotio
- Using Reverse Proxy: no
- Browser: not related
```

Nightly

### Anything else?
[prowlarr.trace.txt](https://github.com/Prowlarr/Prowlarr/files/6945247/prowlarr.trace.txt)

[AB#1405](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/1405)

Nyuels:
#### Database Migration
NO

#### Description
Internet Archive might return results with some missing fields - this PR will solve those cases in the following way:

- if the InfoHash (btih) is missing: don't try to build and set a magnet url
- if the mediatype is missing (should not happen/never happened to me): assign "Other" category

#### Todos
- [ ] Tests

* Fixes #401

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/InternetArchive.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/InternetArchive.cs
@@ -219,7 +219,7 @@ namespace NzbDrone.Core.Indexers.Definitions
                 var downloadUrl = string.Format("{0}/download/{1}/{1}_archive.torrent", _settings.BaseUrl.TrimEnd('/'), searchResult.Identifier);
                 var detailsUrl = string.Format("{0}/details/{1}", _settings.BaseUrl.TrimEnd('/'), searchResult.Identifier);
 
-                var category = _categories.MapTrackerCatToNewznab(searchResult.MediaType);
+                var category = _categories.MapTrackerCatToNewznab(searchResult.MediaType ?? "other");
 
                 var release = new TorrentInfo
                 {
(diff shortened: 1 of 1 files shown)


bakerboy448 (approved):
seems reasonable

Qstick:
Why are we not setting the download link as Jackett does if there is a torrent instead of an infohash to set the magnet?

From Jackett:
`var link = new Uri(LinkUrl + id + "/" + id + "_archive.torrent");`

This would equate to:
`DownloadUrl = LinkUrl + id + "/" + id + "_archive.torrent"`

Nyuels:
> Why are we not setting the download link as Jackett does if there is a torrent instead of an infohash to set the magnet?

We do set both actually - the download URL is constructed in [InternetArchive.cs#L221](https://github.com/Prowlarr/Prowlarr/blob/d5b34e8c038f140627b807a40552e76d5d17d9a1/src/NzbDrone.Core/Indexers/Definitions/InternetArchive.cs#L221) and set in [InternetArchive.cs#L230](https://github.com/Prowlarr/Prowlarr/blob/d5b34e8c038f140627b807a40552e76d5d17d9a1/src/NzbDrone.Core/Indexers/Definitions/InternetArchive.cs#L230).

Issue #394, opened 2021-08-04
[FEAT]: Add (App Sync) Support for Indexers who only have no category / assumed category 8000 (i.e. Other)


bakerboy448:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

Certain indexers - such as BTDigg - do not support categories and default to `8000`

We should add support for these types of indexers syncing to the apps and also need to add a method to guess/assume a category for the results returned.  Possibly via regexing the release name

### Describe the solution you'd like

Discussed above

### Describe alternatives you've considered

Not supporting these indexers in apps at all

### Anything else?

N/A

[AB#1385](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/1385)

toxic0berliner:
as you can see in the many linked issues, not all default to 8000 categories. Some will need keywordless searches and will return some non-standard category like Torrent9 which says this : 
````
Torrent9 only returns category Movies in its Keywordless search results page.
To pass your apps' indexer TEST you will need to include the 131681(Movies) category.
````
So this feature request should also hopefully provide some way to add 131681 or something else when setting up an app for this indexer so it adds these categories.

Qstick:
This can already be done by adjusting sync categories under advanced settings for your apps to include `Other`. 

For the case of torrent9, their `Movies` gets mapped to Torznab Movies (2000). Thus when selecting that it should ensure that Torrent9 is brought in based on the yml mapping.

toxic0berliner:
Thanks for the tip but I fear this didn't work.
I added Others in sonarr's settings in prowlarr but I still get this:
````
	8:27am	ApplicationService	NzbDrone.Core.Applications.ApplicationService HTTP request failed: [400:BadRequest] [POST] at [http://sonarr:8989/api/v3/indexer]	
8:27am	HttpClient	HTTP Error - Res: [POST] http://sonarr:8989/api/v3/indexer: 400.BadRequest [ { "isWarning": false, "propertyName": "", "errorMessage": "Query successful, but no results were returned from your indexer. This may be an issue with the indexer or your indexer category settings.", "severity": "error" } ]	
8:27am	Cardigann	Adding request: https://www.torrent9.pw/top	
````
Other indexers get added to sonarr without issues but it seems this one doesn't pass the sonarr test.
Any help will be very much appreciated.
Thanks in advance.

[...]

PR #423, opened 2021-08-15
Update: (Indexer) BinSearch to support RSS and basic group guessing.

Nyuels:
#### Database Migration
NO

#### Description
This PR introduces RSS/empty searches to the BinSearch Indexer.
To allow this, two new advanced settings are added which make use of BinSearch's [group browsing](https://binsearch.info/browse.php):

- `Fallback Newsgroup`: specifies the group to browse in case of RSS/empty searches
- `Fallback Server`: specifies the backend server BinSearch will use to browse the `Fallback Newsgroup` (_Primary_ indexes the most popular groups, _Secondary_ other groups)

In addition, this PR would add simple category guessing based on the list of filetypes BinSearch returns for each result entry.

#### Todos
- [ ] Extend the basic category guessing.
- [ ] Split PR into the two features?


src/NzbDrone.Core/Indexers/Definitions/BinSearch.cs
```diff
@@ -52,7 +56,7 @@ private IndexerCapabilities SetCapabilities()
             {
                 TvSearchParams = new List<TvSearchParam>
                                    {
-                                       TvSearchParam.Q, TvSearchParam.Season, TvSearchParam.Ep
+                                       TvSearchParam.Q
```
Qstick, line 59: We should keep the Season and Ep as we use it to concat the search term for TV

src/NzbDrone.Core/Indexers/Definitions/BinSearch.cs
```diff
@@ -165,65 +198,141 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
...
+                    .Select(textNode => FileTypesRegex.Match(textNode.TextContent))
+                    .Where(textMatch => textMatch.Success)
+                    .Select(fileTypeMatch => fileTypeMatch.Groups["filetype"].Value);
+
+                if (fileTypes.Count() > 0)
+                {
+                    // Guess category based on filetype
+                    if (fileTypes.Any(fileType => CommonVideoFileTypes.Contains(fileType)))
+                    {
+                        release.Categories = new List<IndexerCategory> { NewznabStandardCategory.Movies, NewznabStandardCategory.TV, NewznabStandardCategory.XXX };
```
Qstick, line 267: This is problematic, we don't want to tag everything with XXX

src/NzbDrone.Core/Indexers/Definitions/BinSearch.cs
```diff
@@ -165,65 +198,141 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
...
+        private static readonly BinSearchSettingsValidator Validator = new BinSearchSettingsValidator();
+
         [FieldDefinition(1, Label = "Base Url", Type = FieldType.Select, SelectOptionsProviderAction = "getUrls", HelpText = "Select which baseurl Prowlarr will use for requests to the site")]
         public string BaseUrl { get; set; }
 
-        [FieldDefinition(2)]
+        [FieldDefinition(2, Label = "Fallback Newsgroup", Advanced = true, Type = FieldType.Textbox, HelpText = "Use this newsgroup for empty searches.")]
+        public string FallbackGroup { get; set; }
+
+        [FieldDefinition(3, Label = "Fallback Server", Advanced = true, Type = FieldType.Select, SelectOptions = typeof(BinSearchFallbackServer), HelpText = "Use this Binsearch server for empty searches.")]
```
Qstick, line 315: Is this really needed? What's the difference when selecting Primary or Secondary?

src/NzbDrone.Core/Indexers/Definitions/BinSearch.cs
```diff
@@ -165,65 +198,141 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
...
-        private static readonly Regex ParseSizeRegex = new Regex(@"size: (?<size>[0-9]+(\.[0-9]+)?).(?<unit>(GB|MB|KB|B))");
-
         public Action<IDictionary<string, string>, DateTime?> CookiesUpdater { get; set; }
     }
 
+    public class BinSearchSettingsValidator : AbstractValidator<BinSearchSettings>
+    {
+        public BinSearchSettingsValidator()
+        {
+            RuleFor(c => c.BaseUrl).NotEmpty();
```
Qstick, line 298: We default to the first URL when empty, so this really isn't needed.

Qstick (changes requested):

src/NzbDrone.Core/Indexers/Definitions/BinSearch.cs
Nyuels: I haven't checked all, but from comparisons I deducted that the list of indexed groups for [Primary](https://binsearch.info/groups.php) and [Secondary](https://binsearch.info/groups.php?server=2) is disjunct:
Primary "contains" the most popular groups, while Secondary "contains" the remaining ones, BinSearch indexes.

src/NzbDrone.Core/Indexers/Definitions/BinSearch.cs
Nyuels: Would Movies and TV category be okay though?
Otherwise it is probably better to wait for a uniform implementation (https://github.com/Prowlarr/Prowlarr/issues/394).

Qstick: That's probably fine for now.

Qstick:
Feel free to bring this out of draft when ready

Nyuels:
> Feel free to bring this out of draft when ready

Will do, I wasn't quite happy about the category tagging, so I tried to make it more reusable.
It would probably be a good idea to update/overhaul/expand on that in the future, depending on needs.

Nyuels:
Updated the category guessing to also include special token in the release title.
By doing so I moved both Parsers into Prowlarr.Core.Parser.Category for future reusability (e.g. https://github.com/Prowlarr/Prowlarr/issues/394).

Qstick:
> Updated the category guessing to also include special token in the release title.
> By doing so I moved both Parsers into Prowlarr.Core.Parser.Category for future reusability (e.g. #394).

Nice work 👍 I’ll have a go

Issue #437, opened 2021-08-18
[BUG]: HDBits returning results when it shouldn't?


stevezau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When searching in sonarr for something that I know does not exist in a hdbits, it is returning results for some reason. 

Debugging sonarr, it is calling Prowlarr via api
```
https://hidden.url/3/api?t=tvsearch&extended=1&apikey={removed}&q=soemthingthatdoesnotexistshouldnotreturn
``` 

Logs from Prowlarr
```
2021-08-18 18:26:39.9|Trace|Http|Req: 2424 [GET] /3/api?t=tvsearch&extended=1&apikey=(removed)&q=soemthingthatdoesnotexistshouldnotreturn (from 14.*.*.20)
2021-08-18 18:26:39.9|Info|NzbSearchService|Searching indexers: [HDBits] for {Term: soemthingthatdoesnotexistshouldnotreturn, Offset: 0, Limit: 0, Categories: []}
2021-08-18 18:26:39.9|Debug|HDBits|Downloading Feed https://hdbits.org/api/torrents
2021-08-18 18:26:39.9|Trace|ConfigService|Using default config value for 'logindexerresponse' defaultValue:'False'
2021-08-18 18:26:39.9|Trace|IndexerHttpClient|Req: [POST] https://hdbits.org/api/torrents
2021-08-18 18:26:39.9|Trace|ConfigService|Using default config value for 'proxyenabled' defaultValue:'False'
2021-08-18 18:26:40.9|Trace|IndexerHttpClient|Res: [POST] https://hdbits.org/api/torrents: 200.OK (1023 ms)
2021-08-18 18:26:40.9|Trace|EventAggregator|Publishing IndexerQueryEvent
2021-08-18 18:26:40.9|Trace|EventAggregator|IndexerQueryEvent -> HistoryService
2021-08-18 18:26:40.9|Trace|EventAggregator|IndexerQueryEvent <- HistoryService
2021-08-18 18:26:40.9|Debug|NzbSearchService|Total of 50 reports were found for {Term: soemthingthatdoesnotexistshouldnotreturn, Offset: 0, Limit: 0, Categories: []} from 1 indexers
2021-08-18 18:26:40.9|Trace|Http|Res: 2424 [GET] /3/api?t=tvsearch&extended=1&apikey=(removed)&q=soemthingthatdoesnotexistshouldnotreturn: 200.OK (1048 ms)
[...]

PR #444, opened 2021-08-21
Fixing HDBits not parsing the search term

stevezau:
#### Database Migration
NO

#### Description
When sonarr searches for a tvshow, sometimes the tvdb might be null. The code today checks if tvdb == 0.. which means the search term is never applied as tvdb is null.

This fixed handles the searching properly.

* Fixes https://github.com/Prowlarr/Prowlarr/issues/437


stevezau:
Tested and works locally. IMDB didn't need fixing

Qstick (approved):

PR #454, opened 2021-08-22
HDBits to parse IMDB using parser utils

stevezau:
#### Database Migration
No

#### Description
Following from https://github.com/Prowlarr/Prowlarr/pull/444. As per @Qstick's request to also fix imdb to use the parser utils.


Qstick:
Doesn't HDBits support IMDB search for TV as well?

stevezau:
@Qstick you're right.. updated it to handle imdb in tvsearch

PR #486, opened 2021-09-04
Update Indexers to Indexer Proxies

stevietv:
the settings submenu said "Indexers" whereas this is actually the settings for Indexer Proxies - updating the translate link to correct this

#### Database Migration
NO

#### Description
Clarifying the menue for a better user experience

#### Todos
- [ ] Tests
- [X] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)

### DIFF
--- a/frontend/src/Settings/Settings.js
+++ b/frontend/src/Settings/Settings.js
@@ -18,7 +18,7 @@ function Settings() {
           className={styles.link}
           to="/settings/indexers"
         >
-          {translate('Indexers')}
+          {translate('IndexerProxies')}
         </Link>
 
         <div className={styles.summary}>


bakerboy448:
Based on previous discussions with @Qstick 

This will be for more than just proxies long term.

thanks for the PR though :)

stevietv:
@bakerboy448 no problem! I must say that I puzzled for 10 minutes on where to add indexers as it seems I could only add indexer proxies! Turns out you don't add them in the settings at all, of course! But coming from a Sonarr/Radarr world where indexers are in the settings that was where I was looking.

Indexer Settings is then maybe a better name than just Indexers?

PR #607, opened 2021-11-16
Add a filter bar to the "Add Indexers" modal

### LINKED ISSUE #94: more sorting options and filters for the "add indexer" list
more fields to sort indexers by would be very helpful. things like language and available categories would be very useful to be added. it would also help if you could filter by all these fields. jackett recently added filtering and it was a great addition

[AB#38](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/38)

csandman:
**EDIT:** I just noticed #178 existed, this is realistically a duplicate of that.  It looks like that one is further along too so I can close this if its unnecessary.

#### Database Migration
NO

#### Description
This PR adds a filter bar to the "Add Indexer" modal, which gives options for filtering the list by Protocol, Language, and Privacy Level.

This is my first PR to a *arr project, so I'm not sure if my coding style matches the rest of what the project is expecting. I am a React developer by trade, but I haven't used class components in a couple years. If you want to make big changes to how I did this, or just nix it and build this feature up from scratch, by all means go for it. The way I implemented it, the languages and privacy types are just built up from the list of indexers coming back, and I'm not sure if it would be better to pull those lists from the backend directly but .net is a little outside my scope.

So please, tear this apart and I'll try and fix it!

image

image

#### Todos
- [ ] Figure out how to fit the filters and table in a mobile layout (perhaps move into the scroller?)
- [ ] Change coding style (component structure and whatnot) to match the rest of the project?

* Closes #94

### DIFF
--- a/frontend/src/Indexer/Add/AddIndexerModalContent.css
+++ b/frontend/src/Indexer/Add/AddIndexerModalContent.css
@@ -16,7 +16,7 @@
   composes: input from '~Components/Form/TextInput.css';
 
   flex: 0 0 auto;
-  margin-bottom: 20px;
+  margin-bottom: 16px;
 }
 
 .alert {
@@ -27,4 +27,43 @@
 
 .scroller {
   flex: 1 1 auto;
+}
+
+.filterRow {
+  display: flex;
+  margin-bottom: 20px;
+}
+
+.filterContainer {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  flex: 1;
+  margin-right: 12px;
+}
+
+.filterContainer:last-child {
+  margin-right: 0;
+}
+
+.filterLabel {
+  font-weight: bold;
+  margin-bottom: 3px;
+}
+
+@media only screen and (max-width: $breakpointSmall) {
+  .filterRow {
+    flex-direction: column;
+  }
+
+  .filterContainer {
+    margin-right: 0;
+    margin-bottom: 12px;
+  }
+
+  .scroller {
+    margin-left: -30px;
+    margin-right: -30px;
+    margin-bottom: -30px;
+  }
 }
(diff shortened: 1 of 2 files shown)


Qstick:
Thanks, let's keep this open. The other switches everything to functional components which we aren't excited about

csandman:
Ok will do! To be honest, I much prefer functional components myself, but I tried to stick with the style of the project.

Qstick:
Yea we'll likely have to start switching at some point, it's just what we've been used to.

Qstick:
Should we keep the info blurb above the filters? Also, seems like we don't have much space on mobile with both the blurb and filters. Maybe a later fix is to autohide something

csandman:
Yeah I was trying to think about the best way to make it fit on mobile, removing the blurb would definitely help the most.  Or we could just hide it specifically on mobile?

Qstick:
Lets do that for now until we get a better way to dismiss it

frontend/src/Indexer/Add/AddIndexerModalContent.css
```diff
@@ -28,3 +28,42 @@
...
+    margin-right: 0;
+    margin-bottom: 12px;
+  }
+
+  .scroller {
+    margin-right: -30px;
+    margin-bottom: -30px;
+    margin-left: -30px;
+  }
+}
```
Qstick, line 69: Add mobile override for `alert` class with `display: none` as discussed. Then good.

Qstick (changes requested):

frontend/src/Indexer/Add/AddIndexerModalContent.css
csandman: Ok, just added it in my most recent push

csandman:
I hid the alert on smaller screen sizes and this is what it looks like on my phone:

image

PR #617, opened 2021-11-18
Net6.0

Qstick:
#### Database Migration
NO

#### Description
Net6 Build and new packages

#### Todos
- [ ] Tests


Qstick:
This should be good for now, pulls over the net6, package and proxy work from Sonarr/Readarr without the new http stack which is causing issues.

bakerboy448:
Wiki updates? ;)
Or lazy servarr links so I can easily and lazily update the wiki at least

PR #771, opened 2022-01-11
New: (MoreThanTV) Add MoreThanTV

DevYukine:
#### Database Migration
NO

#### Description
Adds MoreThanTV C# indexer mostly taken from [Jackett](https://github.com/Jackett/Jackett/blob/master/src/Jackett.Common/Indexers/MoreThanTV.cs).

The implementation is as great as it can get right now, some stuff isn't working super well due to how the site handles things (when Torrents are grouped it will show the combined stats for Size/Seeders/Leechers etc for all of them so we have no way knowing the actual information for a single torrent of the group and just use the grouped info for all torrents) but that's kinda rare so overall it should be good to go until there API is ready.

#### Todos
- [x] Tests

None

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/MoreThanTV.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/MoreThanTV.cs
@@ -328,7 +328,7 @@ public class MoreThanTVSettings : IIndexerSettings
     [FieldDefinition(1, Label = "Base Url", Type = FieldType.Select, SelectOptionsProviderAction = "getUrls", HelpText = "Select which baseurl Prowlarr will use for requests to the site")]
     public string BaseUrl { get; set; }
 
-    [FieldDefinition(2, Label = "Cookie", HelpText = "Enter the cookie for the site", HelpLink = "https://wiki.servarr.com/prowlarr/faq#finding-cookies")]
+    [FieldDefinition(2, Label = "Cookie", HelpText = "Enter the cookies for the site, just copy everything after 'cookie:' from the request headers to the site", HelpLink = "https://wiki.servarr.com/prowlarr/faq#finding-cookies")]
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Parser/ParseUtil.cs
```diff
@@ -113,8 +113,16 @@ public static string GetArgumentFromQueryString(string url, string argument)
 
         public static long GetBytes(string str)
         {
+            var valStr = new string(str.Where(c => char.IsDigit(c) || c == '.' || c == ',').ToArray());
```
bakerboy448, line 116: We already handle this functionality with `NormalizeNumber`

DevYukine: I reverted my change since i just saw its already implemted 😄

src/NzbDrone.Core/Indexers/Definitions/MoreThanTV.cs
```diff
@@ -0,0 +1,341 @@
...
+        Cookie = "";
+    }
+
+    [FieldDefinition(1, Label = "Base Url", Type = FieldType.Select, SelectOptionsProviderAction = "getUrls", HelpText = "Select which baseurl Prowlarr will use for requests to the site")]
+    public string BaseUrl { get; set; }
+
+    [FieldDefinition(2, Label = "Cookie", HelpText = "Enter the cookie for the site", HelpLink = "https://wiki.servarr.com/prowlarr/faq#finding-cookies")]
+    public string Cookie { get; set; }
+
+    [FieldDefinition(5)]
```
Qstick, line 334: 1,2, 5?

DevYukine: Updated 👌

PR #823, opened 2022-02-05
Def rework

Qstick:
#### Database Migration
YES - 16

#### Description
Switch all indexers over to base implementations and use yml to define caps, links, categories, etc. 

#### Todos
- [ ] Tests
- [ ] Ymls
- [ ] Migration for non-cardigann


bakerboy448:
Fixes https://github.com/Prowlarr/Prowlarr/issues/938

PR #984, opened 2022-05-02
Fixed: (AnimeBytes) Handle series synonyms with commas

lflare:
#### Database Migration
NO

#### Description
A short attempt at fixing Prowlarr's handling of AnimeBytes `synonyms` field, when it failed to parse `Spy x Family` for me.

Example payload:
```json
"Synonymns":{"2":"SPYxFAMILY, Spy x Family"}
```

#### Todos
- Tests

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
@@ -228,10 +228,18 @@ namespace NzbDrone.Core.Indexers.Definitions
                         if (syn.StringArray != null)
                         {
                             synonyms.AddRange(syn.StringArray);
+                            if (syn.StringArray.Count >= 3)
+                            {
+                                synonyms.AddRange(syn.StringArray[2].Split(',').Select(t => t.Trim()));
+                            }
                         }
                         else
                         {
                             synonyms.AddRange(syn.StringMap.Values);
+                            if (syn.StringMap.ContainsKey("2"))
+                            {
+                                synonyms.AddRange(syn.StringMap["2"].Split(',').Select(t => t.Trim()));
+                            }
                         }
                     }
 
--- a/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
@@ -227,15 +227,33 @@ namespace NzbDrone.Core.Indexers.Definitions
 
                         if (syn.StringArray != null)
                         {
-                            synonyms.AddRange(syn.StringArray);
-                            if (syn.StringArray.Count >= 3)
+                            if (syn.StringArray.Count >= 1)
+                            {
+                                synonyms.Add(syn.StringArray[0]);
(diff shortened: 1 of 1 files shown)


bakerboy448:
@DevYukine

giosann:
https://github.com/Jackett/Jackett/issues/8683
this was the original request, I still don't think that there is a way to get the filenames.

lflare:
> [Jackett/Jackett#8683](https://github.com/Jackett/Jackett/issues/8683) this was the original request, I still don't think that there is a way to get the filenames.

Theoretically, Prowlarr can pull the `.torrent` file, parse it and check the filenames, but that's a functionality far beyond what we currently have supported. It would be an interesting feature to have, but for now, properly handling alternative titles split by commas is the next best thing. Also a low-hanging fruit.

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
```diff
@@ -228,10 +228,18 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                         if (syn.StringArray != null)
                         {
                             synonyms.AddRange(syn.StringArray);
+                            if (syn.StringArray.Count >= 3)
```
DevYukine, line 231: This new implementation will double add the "splitted" alias because we first add all aliases as is (e.g. if the array is ["this is an example", "this is an example 2, "aliases, comma seperated"] then you add both the complete comma seperated one and the splitted one)

Instead of doing this you should just add index 0 & 1 and then handle everything from 2 onwards with a split by comma.

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
```diff
@@ -228,10 +228,18 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
...
+                            {
+                                synonyms.AddRange(syn.StringArray[2].Split(',').Select(t => t.Trim()));
+                            }
                         }
                         else
                         {
                             synonyms.AddRange(syn.StringMap.Values);
+                            if (syn.StringMap.ContainsKey("2"))
```
DevYukine, line 239: Same as above, will add the one with key 2 two times instead handling this edge case

DevYukine (changes requested):
You should just do it like Jackett and handle all alternative titles properly instead just one slot and adding everything again if you wanna update this

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
lflare: I've replicated the logic from Jackett. There shouldn't be any synonyms from index 3 onwards, according to the AnimeBytes developers.

lflare:
@DevYukine, please let me know if there's anything else that should be improved upon :)

DevYukine (approved):
Seems fine to me now, matches Jacketts implementation

bakerboy448 (approved):
Looks good to yukine; looks good to me

Issue #232, opened 2021-06-17
Stats Page: Chart Display & Colour Improvements


extrobe:
**Describe the solution you'd like**
The charts on the stats page are really useful, but a couple of 'best practice' changes would make them a little more readable.

1) the charts that are side-by-side both show the same categorical data, but does so in a different order (as each is ranked independently from highest to lowest)

Normal practice would be to show these with the categories in the same order. (See image for proposed before/after).

This tends to make the data far more readable for the user.

![image](https://user-images.githubusercontent.com/25456852/122355155-08399400-cf95-11eb-8d72-ed0fcbf5fecf.png)

2) Charts where you're using multiple legends (eg the Indexer Query counts split between RSS/Auth/Search queries) should use more contrasting colours to make the distinction clearer. (Consider also colour blindness)

3) More minor, but for charts where there aren't multiple legends, don't change the shade across the categories, they should ideally stay the same rather than shading across the categories.

[AB#10](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/10)

TheCatLady:
I personally prefer the data to be sorted by value, since I'm typically interested in seeing which indexers are performing well and which are failing.

Qstick:
I agree that sorting by value isn't an abnormal way to sort, and it's what many care about. So not sure we'll be changing that.

I'm all for updating the stats color scheme as it was mostly a placeholder to get something in place. Welcome to see any suggestions for color options

2fst4u:
As a red-green colourblind user, for colour schemes may I suggest completely disparate hues. You'll never get a scheme that's right for everyone, and making the colours really wild doesn't always look the prettiest, but if you went with for example a yellow, a red and a blue for those ones where it's showing seperate data for a single line it would be easier to read.

For the graphs that show just have one set of data for each bar the colours don't really need to be too disparate because there's nothing blending together and the colour differences don't need to show anything, they're just for show. They're fine as they are. It would be good if the gradient changed depending on the value though, like the most latent indexer being more red or even purple, and the best being a light green.

I'm happy to look over colour suggestions when someone gets around to this.

twolaw:
I suggest to display completely different colors instead of this 50 shades of yellow (search queries vs RSS queries are impossible to distinguish)
As an example :
![image](https://user-images.githubusercontent.com/10365746/166143584-27a8f4a8-0886-46dc-9881-e49dbacb324f.png)

PR #1003, opened 2022-05-14
Stats : diversify chartcolors for doughnut & stackedbar

twolaw:
#### Database Migration
NO

#### Description

On the Stats page, doughnut charts and stackedbar charts are currently unreadable because of the almost same yellow color for all items.
This PR brings a full range of differents colors for the items of those kind of charts.

image

Bar charts are not changed, they remain yellow with all bars the same color (maybe differents colors could be added in the futur when the order of items will be identical in all bar charts).

#### Issue improved by this PR (but not totally fixed yet)
#232

### DIFF
--- a/frontend/src/Components/Chart/DoughnutChart.js
+++ b/frontend/src/Components/Chart/DoughnutChart.js
@@ -29,7 +29,7 @@ class DoughnutChart extends Component {
         datasets: [{
           label: this.props.title,
           data: this.props.data.map((d) => d.value),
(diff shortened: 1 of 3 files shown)


Qstick:
Can we just make a separate color array for this, the bar charts are different shades for each indexer (although not very different to your point), but your PR changes to use the first item in array making them all the same color.

twolaw:
Ok, I updated with a separate colors array.

PR #1063, opened 2022-07-13
New: Notifiarr moved from webhook to API

bakerboy448:
#### Database Migration
NO

#### Description
Pending Notifiarr changes

#### Todos
- Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
+++ b/src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
@@ -1,9 +1,10 @@
-﻿using System;
+using System;
 using System.Collections.Specialized;
-using System.Net;
 using FluentValidation.Results;
 using NLog;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
+using NzbDrone.Core.Configuration;
 
 namespace NzbDrone.Core.Notifications.Notifiarr
 {
@@ -15,27 +16,21 @@ namespace NzbDrone.Core.Notifications.Notifiarr
 
     public class NotifiarrProxy : INotifiarrProxy
     {
-        private const string URL = "https://notifiarr.com/notifier.php";
+        private const string URL = "https://notifiarr.com";
         private readonly IHttpClient _httpClient;
+        private readonly IConfigFileProvider _configFileProvider;
         private readonly Logger _logger;
 
-        public NotifiarrProxy(IHttpClient httpClient, Logger logger)
+        public NotifiarrProxy(IHttpClient httpClient, IConfigFileProvider configFileProvider, Logger logger)
         {
             _httpClient = httpClient;
+            _configFileProvider = configFileProvider;
             _logger = logger;
         }
 
         public void SendNotification(StringDictionary message, NotifiarrSettings settings)
         {
-            try
-            {
                 ProcessNotification(message, settings);
-            }
-            catch (NotifiarrException ex)
-            {
-                _logger.Error(ex, "Unable to send notification");
-                throw new NotifiarrException("Unable to send notification");
-            }
         }
 
         public ValidationFailure Test(NotifiarrSettings settings)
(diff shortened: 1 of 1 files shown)


Qstick:
Can we streamline the error handling in this PR as well? Thanks

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -48,30 +51,24 @@ public ValidationFailure Test(NotifiarrSettings settings)
...
         }
 
         private void ProcessNotification(StringDictionary message, NotifiarrSettings settings)
         {
             try
             {
-                var requestBuilder = new HttpRequestBuilder(URL).Post();
-                requestBuilder.AddFormParameter("api", settings.APIKey).Build();
+                var instanceName = _configFileProvider.InstanceName;
+                var requestBuilder = new HttpRequestBuilder(URL + "/api/v1/notification/prowlarr/" + settings.APIKey).Post();
```
davidnewhall, line 70: @austinwbest wont agree, but these should use `x-api-key` headers to keep the api key out of logs/urls.

bakerboy448: it's cleansed anyway; would only be cleartext logged to console.

davidnewhall: Any upstream proxy can capture the url.

bakerboy448: TBD future problem

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -48,30 +43,25 @@ public ValidationFailure Test(NotifiarrSettings settings)
...
         }
 
         private void ProcessNotification(StringDictionary message, NotifiarrSettings settings)
         {
             try
             {
-                var requestBuilder = new HttpRequestBuilder(URL).Post();
-                requestBuilder.AddFormParameter("api", settings.APIKey).Build();
+                var instanceName = _configFileProvider.InstanceName;
+                var requestBuilder = new HttpRequestBuilder(URL + "/api/v1/notification/prowlarr/").Post();
```
Qstick, line 62: Can we get rid of this trailing `/` or is it needed? Also does this allow us to kill the cleanse regex we added specifically for these?

bakerboy448: @austinwbest @davidnewhall defer to y'all on if the / is needed

Yes can purge the regex then

austinwbest: You can prolly simplify this now since there is no dynamic domains or data and just set this to `https://notifiarr.com/api/v1/notification/prowlarr` and remove the URL variable. Same could be said for the instance name, just set it to `_configFileProvider.InstanceName` explicitly and dump the variable

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -48,30 +43,25 @@ public ValidationFailure Test(NotifiarrSettings settings)
...
+                requestBuilder.AddFormParameter("instanceName", instanceName).Build();
+                requestBuilder.SetHeader("X-API-Key", settings.APIKey);
```
austinwbest, line 64: This looks right but just to be safe, did you fire a test notification from your local build?

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
bakerboy448: no regex kill needed. it was aspecific cleanse added that would also cover generic webhooks

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
bakerboy448: tested works
also confirmed without trailing slash

Qstick (approved):

Issue #120, opened 2021-06-07
Indexer - Cinemageddon - Unable to Add - Redirected to Domain


Junkbite:
**Describe the bug**

cant setup Cinemageddon get `Unable to connect to indexer, check the log for more details`

**To Reproduce**

**Expected behavior**

**Screenshots**
```
Version
0.1.0.370
Package Version
0.1.0.370-ls4 by linuxserver.io
.NET
Yes (5.0.6)
Docker
Yes
DB Migration
7
```

Turn on Trace logs under Settings -> General and wait for the bug to occur again.
**Upload the full log file here (or another site (e.g. pastebin) and link it). Issues will be closed, if they do not include this!**

```
2021-06-06 20:22:04.0|Trace|EventAggregator|ConfigFileSavedEvent <~ ReconfigureLogging
2021-06-06 20:22:04.0|Trace|EventAggregator|HealthCheckCompleteEvent ~> NotificationService
2021-06-06 20:22:04.0|Trace|EventAggregator|ConfigSavedEvent ~> LocalizationService
2021-06-06 20:22:04.0|Trace|EventAggregator|ConfigSavedEvent <~ LocalizationService
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxyenabled' defaultValue:'False'
2021-06-06 20:22:04.0|Trace|EventAggregator|HealthCheckCompleteEvent ~> NotificationService
2021-06-06 20:22:04.0|Trace|EventAggregator|HealthCheckCompleteEvent <~ NotificationService
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxytype' defaultValue:'Http'
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxyhostname' defaultValue:''
2021-06-06 20:22:04.0|Trace|EventAggregator|HealthCheckCompleteEvent <~ NotificationService
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxyport' defaultValue:'8080'
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxyusername' defaultValue:''
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxypassword' defaultValue:''
2021-06-06 20:22:04.0|Trace|ConfigService|Using default config value for 'proxybypassfilter' defaultValue:''
[...]

PR #1121, opened 2022-09-13
Fix mteam fail to add and easily unaviliable issue

### LINKED ISSUE #1061: Can't add indexers m-team (mteam)

### Current Behavior

Add indexer m-team always show "Unable to connect to indexer, check the log for more details"

### Expected Behavior

Indexer added successfully

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Synology DSM 7.1
- Prowlarr:0.4.2.1879
- Docker Install:Yes
- Using Reverse Proxy:No
- Browser:Chrome newest version
```

develop

### Trace Logs?

https://pastebin.com/R321SeGg

luckyboy:
#### Database Migration
NO

#### Description
This PR fix mteam fail to add and easily unavailable issue. 
The site mteam has overloading detection which block access in 120s. So, there are still improvements to do to avoid 
too many access in short period of time.

#### Todos
Http request flow need to improve for decreasing number of requests for login or searching.

* Fixes #1061

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannRequestGenerator.cs
@@ -945,6 +945,11 @@ namespace NzbDrone.Core.Indexers.Cardigann
 
         public bool CheckIfLoginIsNeeded(HttpResponse response)
         {
+            if (_definition.Login == null || _definition.Login.Test == null)
+            {
+                return false;
+            }
+
             if (response.HasHttpRedirect)
             {
                 var domainHint = GetRedirectDomainHint(response);
@@ -956,11 +961,6 @@ namespace NzbDrone.Core.Indexers.Cardigann
                 }
 
                 return true;
-            }
-
-            if (_definition.Login == null || _definition.Login.Test == null)
-            {
-                return false;
             }
 
             if (response.HasHttpError)
@@ -1123,6 +1123,8 @@ namespace NzbDrone.Core.Indexers.Cardigann
                 }
 
                 var request = new CardigannRequest(requestbuilder.SetEncoding(_encoding).Build(), variables, searchPath);
+
+                request.HttpRequest.AllowAutoRedirect = searchPath.Followredirect;
 
                 yield return request;
             }
--- a/src/NzbDrone.Core/Indexers/Definitions/Newznab/NewznabRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Newznab/NewznabRequestGenerator.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
+using DryIoc;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 using NzbDrone.Core.IndexerSearch.Definitions;
@@ -268,7 +269,10 @@ namespace NzbDrone.Core.Indexers.Newznab
                 parameters.Add("offset", searchCriteria.Offset.ToString());
             }
 
-            yield return new IndexerRequest(string.Format("{0}&{1}", baseUrl, parameters.GetQueryString()), HttpAccept.Rss);
+            var request = new IndexerRequest(string.Format("{0}&{1}", baseUrl, parameters.GetQueryString()), HttpAccept.Rss);
+            request.HttpRequest.AllowAutoRedirect = true;
+
+            yield return request;
         }
 
         private static string NewsnabifyTitle(string title)
--- a/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
@@ -355,8 +355,6 @@ namespace NzbDrone.Core.Indexers
                 request.HttpRequest.LogResponseContent = true;
             }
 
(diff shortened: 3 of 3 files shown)


bakerboy448 (changes requested):
- why the build props changes? That seems unrelated to m team?   M1 support already exists and was added with https://github.com/Prowlarr/Prowlarr/commit/c35e257a82dcacdc282124d1fd7d8fefc7cba61f
- What is the purpose of `CFPCheck` and why is it always true?
- What is this new "portal" functionality for Cardigann and how is it to be used?
- Looks like login portal can just leverage the existing login test logic instead, no?
- non-backwards compatible Cardigann changes like this require a Cardigann version bump, a resulting new schema on the definitions, and the corresponding new definition
- believe Mteam was working and fixed with https://github.com/Prowlarr/Prowlarr/pull/1120

luckyboy:
> * why the build props changes? That seems unrelated to m team?   M1 support already exists and was added with [c35e257](https://github.com/Prowlarr/Prowlarr/commit/c35e257a82dcacdc282124d1fd7d8fefc7cba61f)
   
RE:Sorry, this change does not related to mteam, it's a fix for VisualStudio on M1 Mac(RuntimeIdentifier is set to osx-x64 which cause fail to launch program from VS), just ignore it.
> * What is the purpose of `CFPCheck` and why is it always true?
  
 RE:Request to mteam without UserAgent (login flow) will receive a page not found response(html), but CloudFlareDetectionService.IsCloudflareProtected can't handle this scenary well, so, CFPCkeck is purposed to skip this step 
with mteamtp  definition change, and thus make sure no side effect to other indexers.
> * What is this new "portal" functionality for Cardigann and how is it to be used?
   
RE:Portal function is also purposed to fix mteam login checking logic. The original code CardigannRequestGenerator.CheckIfLoginIsNeeded is not strong enough to return a correct login status, which will cause re-login for most searching request, but unfortunately mteam has login security check, too many login will cause account locked, finally the indexer(mteam) will show unavailable. Also CardigannParser.ParseResponse will receive an access blocking page(loading is high) from mteam if too many request in a short peroid of time, so the error handling should also be improved for avoiding return a confused exception in the log file.
> * Looks like login portal can just leverage the existing login test logic instead, no?
    
RE:The reason why login portal is introduced is still related to the login status checking logic. The existing logic check the login status by searching logout.php, but there are pages not contain logout.php(for example the access blocking page), this  will cause re-login.
> * non-backwards compatible Cardigann changes like this require a Cardigann version bump, a resulting new schema on the definitions, and the corresponding new definition
    
RE:There no backwards compatible issue with those code, because these code changes and mteam definition changes just work for mteam only, also there is no side effect without mteam definition change.
> * believe Mteam was working and fixed with #1120
    
RE:#120 just fixed login issue, but mteam indexer still can't work due to lack of proper error handling mentioned above.

bakerboy448:
If the new definition works without error on nightly (other than obviously not working for the known issues) then you don't need to bump Cardigann versions.  If it does - which it looks rather likely. Then you will need to bump the Cardigann version and handle the associated schema update in the indexer repo.

M1 Mac VS build fixes should ideally be in its own branch and ported to all the \*arrs as well


How exactly is the cloudflare protection detection causing issues?

This tracker was already confirmed to be fixed by other users when the redirect logic change was made. Do you have trace logs of it failing on the latest nightly build?

luckyboy:
> If the new definition works without error on nightly (other than obviously not working for the known issues) then you don't need to bump Cardigann versions. If it does - which it looks rather likely. Then you will need to bump the Cardigann version and handle the associated schema update in the indexer repo.
> 
> Pretty sure it'll need the bump.
> 
> M1 Mac VS build fixes should ideally be in its own branch and ported to all the *arrs as well
> 
> How exactly is the cloudflare protection detection causing issues?
> 
> This tracker was already confirmed to be fixed and working by other users when the redirect logic change was made. Do you have trace logs of it failing on the latest nightly build with Flaresolverr 2.2.7 (their latest release) and the FS proxy enabled for this tracker?

Here is the test log from lasted nightly build with Flaresolverr 2.2.7, still failed after login.
https://pastebin.com/iadxbGJ9

By the way, could you please give me an example commit of bump Cardigann versions.?

bakerboy448:
Info logs are not trace logs

luckyboy:
Below is trace and debug files
[prowlarr.debug.txt](https://github.com/Prowlarr/Prowlarr/files/9597088/prowlarr.debug.txt)
[prowlarr.trace.txt](https://github.com/Prowlarr/Prowlarr/files/9597090/prowlarr.trace.txt)

luckyboy:
> Info logs are not trace logs
> 
> Indexers Repo has the versions linked to their prowlarr commits IIRC
Below is debug and trace files when failure happen
[prowlarr.trace.txt](https://github.com/Prowlarr/Prowlarr/files/9597099/prowlarr.trace.txt)
[prowlarr.debug.txt](https://github.com/Prowlarr/Prowlarr/files/9597100/prowlarr.debug.txt)

usnail:
> 



> @Qstick cardigann is still puking on a redirect being returned after Flaresolverr solves it
> 
> ```
> 16:00:37.7|Warn|Cardigann|https://kp.m-team.cc/adult.php?search=Elvis&incldead=0&spstate=0&search_area=0&search_mode=0&sort=4&type=desc
> 
> [v0.4.6.1969] NzbDrone.Core.Indexers.Exceptions.IndexerException: Unexpected response status Redirect code from API request
> ```

some times,site always redirect to verify page even though has login done（login otherwhere）

PR #1215, opened 2022-11-26
Fixed: (Avistaz Family) Age Parsing

### LINKED ISSUE #1212: Update avistaz network age logic

Prowlarr assumes times from sites are UTC which causes issues with age calculation.

avistaz sites per staff are UTC-5;  prowlarr should treat the dates/times accordingly and calculate appropriately.

N/a

### Anything else?

> I am not 100% sure what the exact timezone is, but I think UTC-5 is correct, as that should be approx where the tracker server is located.

[AB#4177](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/4177)

bakerboy448:
#### Database Migration
NO

#### Description
Pre Fix
image
Post Fix
image
image

#### Todos
- [x] Tests

* Fixes #1212

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
@@ -61,7 +61,7 @@ namespace NzbDrone.Core.Indexers.Definitions.Avistaz
                     InfoUrl = details,
                     Guid = details,
                     Categories = cats,
-                    PublishDate = row.CreatedAt,
+                    PublishDate = DateTime.Parse(row.CreatedAt + "-05:00").ToUniversalTime(), // Avistaz does not specify a timezone & returns server time
                     Size = row.FileSize,
                     Files = row.FileCount,
                     Grabs = row.Completed,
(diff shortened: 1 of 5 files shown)


bakerboy448:
tests pass fine - azure being azure....

Qstick:
> tests pass locally - azure failing for alpine for no rationale reason

The rational reason is that the code you added doesn't work on alpine. Likely due to how you implemented the timezone parsing.

src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
```diff
@@ -14,6 +14,9 @@ public class AvistazParser : IParseIndexerResponse
     {
         private readonly HashSet<string> _hdResolutions = new HashSet<string> { "1080p", "1080i", "720p" };
 
+        // Servers are Eastern US Time Zone
+        private readonly TimeZoneInfo _avistazTimeZone = TimeZoneInfo.FindSystemTimeZoneById("US Eastern Standard Time");
```
Qstick, line 18:
```suggestion
        private readonly TimeZoneInfo _avistazTimeZone = TimeZoneInfo.FindSystemTimeZoneById("America/New_York");
```

Use official IANA timezone to ensure this works on more OSs

Qstick:
The other option is we append TZ data to the incoming string we parse given FindSystemTimeZoneById can vary based on on OS culture data. 

We'll see if this works, but sometime in mid-late net6 the time zone discrepancies between OS's were supposed to be resolved so that `FindSystemTimeZoneById ` could parse either on either OS (Windows vs Posix)

bakerboy448:
> The other option is we append TZ data to the incoming string we parse given FindSystemTimeZoneById can vary based on on OS culture data.
> 
> We'll see if this works, but sometime in mid-late net6 the time zone discrepancies between OS's were supposed to be resolved so that `FindSystemTimeZoneById ` could parse either on either OS (Windows vs Posix)

supposedly fixed in the previews
https://devblogs.microsoft.com/dotnet/date-time-and-time-zone-enhancements-in-net-6/

bakerboy448:
BSD now failing due to node `This step requires a node version that does not exist in the agent filesystem. Path: /usr/home/azure/prowlarr/externals/node16/bin/node`

PR #1242, opened 2022-12-18
New: Smarter Newznab category mapping

### LINKED ISSUE #1117: SceneNZBs Custom Categories (German Usenet) Etc

The Problem is, that i can´t use full sync with SceneNZBs because SceneNZBs uses custom Categories, but i can´t select them at the Prowlarr Sync View.
Prowlarr Categories:
image
SceneNZBs Categories:
image

It would be nice if Prowlarr could check if some of the Indexers with Movies for example have custom Categories and add them to the dropdown view. Currently i have to set the indexer for each instance by hand.

Setting it everywhere by hand

### Anything else?

https://scenenzbs.com/

[AB#3711](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/3711)

Qstick:
#### Database Migration
NO

#### Description
Be smarter about mapping nab capabilities. Instead of simply using the id from the indexer and assuming it's going to match one of our standards, lets also compare by name. This allows us to better match those indexers which go by different Id standards without the need of maintaining ymls and such.

#### Todos
- [ ] Tests

* Fixes #1117

### DIFF
--- a/src/NzbDrone.Core.Test/Files/Indexers/Newznab/newznab_caps.xml
+++ b/src/NzbDrone.Core.Test/Files/Indexers/Newznab/newznab_caps.xml
@@ -18,6 +18,7 @@
       <subcat id="5030" name="SD"/>
       <subcat id="5060" name="Sport"/>
       <subcat id="5010" name="WEB-DL"/>
+      <subcat id="5999" name="Other"/>
     </category>
     <category id="7000" name="Other">
       <subcat id="7010" name="Misc"/>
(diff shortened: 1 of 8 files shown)


bakerboy448:
https://github.com/Prowlarr/Prowlarr/issues/1117

Y0ngg4n:
@Qstick so it seems like the SceneNZBs issue is fixed now, but how is it working now. Am i adding SceneNZBs as `Generic Newznab`? How do i select in Prowlarr which categories to sync that are non standard (like with SceneNZBs German categories) with my applications?

PR #1254, opened 2022-12-23
Avistaz response improvements

bakerboy448:
#### Database Migration
NO

#### Description
minor fixups for Avistaz response improvements

#### Todos
- [] Tests
- [] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)
- [] [Wiki Updates](https://wiki.servarr.com)

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
@@ -24,24 +24,26 @@ namespace NzbDrone.Core.Indexers.Definitions.Avistaz
         {
             var torrentInfos = new List<TorrentInfo>();
 
-            if (indexerResponse.HttpResponse.StatusCode == HttpStatusCode.NotFound)
+            if (!indexerResponse.HttpResponse.Headers.ContentType.Contains(HttpAccept.Json.Value))
             {
-                return torrentInfos.ToArray();
-            }
-
-            if (indexerResponse.HttpResponse.StatusCode == HttpStatusCode.TooManyRequests)
-            {
-                throw new RequestLimitReachedException(indexerResponse, "API Request Limit Reached");
+                throw new IndexerException(indexerResponse, $"Unexpected response header {indexerResponse.HttpResponse.Headers.ContentType} from API request, expected {HttpAccept.Json.Value}");
             }
 
             if (indexerResponse.HttpResponse.StatusCode != HttpStatusCode.OK)
             {
-                throw new IndexerException(indexerResponse, $"Unexpected response status {indexerResponse.HttpResponse.StatusCode} code from API request");
-            }
+                if (indexerResponse.HttpResponse.StatusCode == HttpStatusCode.NotFound)
+                {
+                    // No results found
+                    return torrentInfos.ToArray();
+                }
 
-            if (!indexerResponse.HttpResponse.Headers.ContentType.Contains(HttpAccept.Json.Value))
-            {
-                throw new IndexerException(indexerResponse, $"Unexpected response header {indexerResponse.HttpResponse.Headers.ContentType} from API request, expected {HttpAccept.Json.Value}");
+                HttpResponse<AvistazErrorResponse> jsonErrorResponse = new HttpResponse<AvistazErrorResponse>(indexerResponse.HttpResponse);
+                if (indexerResponse.HttpResponse.StatusCode == HttpStatusCode.Unauthorized)
+                {
+                    throw new IndexerAuthException(string.Empty, jsonErrorResponse.Resource?.Message ?? "Unauthorized request to indexer");
+                }
+
+                throw new IndexerException(indexerResponse, $"Unexpected response status {indexerResponse.HttpResponse.StatusCode} code from API request");
             }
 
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
```diff
@@ -24,24 +24,30 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
         {
             var torrentInfos = new List<TorrentInfo>();
 
+            if (indexerResponse.HttpResponse.StatusCode == HttpStatusCode.TooManyRequests)
+            {
+                throw new TooManyRequestsException(indexerResponse.HttpRequest, indexerResponse.HttpResponse, TimeSpan.FromHours(1));
```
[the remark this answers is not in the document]
bakerboy448: what's the vlaues of `x-ratelimit-remaining`

src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
```diff
@@ -24,24 +24,31 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
...
             {
-                throw new RequestLimitReachedException(indexerResponse, "API Request Limit Reached");
+                throw new IndexerException(indexerResponse, $"Unexpected response header {indexerResponse.HttpResponse.Headers.ContentType} from API request, expected {HttpAccept.Json.Value}");
             }
 
             if (indexerResponse.HttpResponse.StatusCode != HttpStatusCode.OK)
             {
-                throw new IndexerException(indexerResponse, $"Unexpected response status {indexerResponse.HttpResponse.StatusCode} code from API request");
-            }
+                if (indexerResponse.HttpResponse.StatusCode == HttpStatusCode.TooManyRequests)
```
Qstick, line 34: This actually gets caught in HttpIndexerBase globally before we even go to parsing. No use in having this in an individual indexer unless the indexer is using something other than the correct status code to indicate Request limit

https://github.com/Prowlarr/Prowlarr/blob/eff09c1f72a91226d1142cd8d1ca51ab0dedc88d/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs#L382-L390

Qstick (changes requested):

src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
Qstick: In that case, this change won't do anything to fix it. Looking at how the exception is setup. Regarding the case sensitivity, it should be case-insensitive, but we'd need to run a few tests to see. 

The way it's setup, it'll never actually retry this same request again, it'll just shut the indexer down for the `Retry-After` time.

Qstick: A quick test seems to indicate case shouldn't matter either way we grab. 

```
        [TestCase("retry-after")]
        [TestCase("Retry-After")]
        public void should_get_header_by_either_case(string headerName)
        {
            var message = new HttpResponseMessage();
            var headers = message.Headers;
            headers.Add(headerName, "5");

            var httpheader = new HttpHeader(headers.ToNameValueCollection());

            httpheader["Retry-After"].Should().Be("5");
            httpheader.Get("Retry-After").Should().Be("5");
        }
```

image

bakerboy448: On a tangential note, does prowlarr pass the retry time backoff to the Torznab/newznab response and do Starrs properly handle it?

Qstick: > I understand the logic, I want to be disabled for maximum 60s and not one hour.
> 
> I just debugged what headers the response in the exception have and are these: `Date, Transfer-Encoding, Connection, Cache-Control, X-XSS-Protection, X-Content-Type-Options, Referrer-Policy, Content-Security-Policy, permissions-policy, Strict-Transport-Security, CF-Cache-Status, Report-To, NEL, Server, CF-RAY, Content-Type`. There is no `Retry-After`.

Ah, so we are losing it somewhere it seems?

Qstick: Avistaz drops it when we request content type `application/json`. That's the issue

No `Accept` header
image

Accept Header with Json
image

Qstick: Yea, I let our contact on their staff know on Discord, though I doubt they will do anything with it quickly. We'll have to work around it for now. 

I think that's our backoff logic, It's identical to Sonarr/Radarr so it may need some adjustments for fine tuning on Prowlarr. Currently there are escalation levels, it uses the level that represents the best match for the minimumBackoff that's passed in. The lowest level is 5 min.

Qstick:
What are you attempting to fix with BTN? `TooManyRequestException` defaults to 1 hour (same as you are passing) and `RequestLimitReachedException` also defaults to 1 hour. So it seems like you aren't changing anything here.

bakerboy448:
reworked - Avistaz response improvements

PR #1255, opened 2022-12-25
New: (Indexer) Torrent Libble with 2FA support

mynameisbogdan:
#### Database Migration
NO

#### Description
Add Libble as C# indexer with 2FA support

Please don't hesitate to do a code review. 🤒


Qstick:
I've no real way to test this one, so I'm good with it going in when you are comfortable. I'll do a quick code scan tonight.

PR #1262, opened 2022-12-27
Fixed: (Indexer) Changed OPS to use filters for categories, label and year

mynameisbogdan:
#### Database Migration
 NO

#### Description
- Currently when filtering by something different than `Audio`, for example `Books`, even if the API returns something it shows 0 rows since the latest aren't eligible as books. This should fix that. 
- Added `Label` and `Year` to search.


Qstick:
Nice improvements here, I'll have a quick review and get it in later today.

Qstick (approved):

PR #1275, opened 2023-01-01
Fixed: (Avistaz) Workaround for fetching "retry-after" header 

mynameisbogdan:
not present when using "Accept: application/json"

#### Database Migration
NO


Qstick (approved):
Thanks for all the help as of late. Much appreciated!

PR #1283, opened 2023-01-03
Fixed: (Filelist) Use UTC in tests

mynameisbogdan:
#### Database Migration
NO


Qstick (approved):
Thanks for knocking this out, was planning to do it when I got home so this was a nice surprise :D

PR #1290, opened 2023-01-05
Fixed: (Cardigann) Apply RateLimit by using RequestDelay from definitions

### LINKED ISSUE #1287: [Cardigann] "requestDelay" not supported?

### Current Behavior

[Wiki](https://wiki.servarr.com/prowlarr/cardigann-yml-definition#header) says that it supports `requestDelay`, but I didn't find anything in the code to back this affirmation up. 

This it's a little tricky because it's necessary with some trackers that rate limit you even on browsing/scraping, not necessary just on API use.

### Expected Behavior

`requestDelay` to be supported.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: unraid
- Prowlarr: 1.0.1.2220
- Docker Install: Y
- Using Reverse Proxy: -
- Browser: -
```

Develop

### Trace Logs?

N/A

[AB#4631](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/4631)

mynameisbogdan:
Just as a safe measure, use `RequestDelay` value only if it's greater than the inherited RateLimit (2s currently).

#### Database Migration
NO

* Fixes #1287

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Cardigann/Cardigann.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Cardigann/Cardigann.cs
@@ -30,8 +30,23 @@ namespace NzbDrone.Core.Indexers.Cardigann
         public override IndexerPrivacy Privacy => IndexerPrivacy.Private;
 
         // Page size is different per indexer, setting to 1 ensures we don't break out of paging logic
-        // thinking its a partial page and insteaad all search_path requests are run for each indexer
+        // thinking its a partial page and instead all search_path requests are run for each indexer
         public override int PageSize => 1;
+
+        public override TimeSpan RateLimit
+        {
+            get
+            {
+                var definition = _definitionService.GetCachedDefinition(Settings.DefinitionFile);
+
+                if (definition.RequestDelay.HasValue && definition.RequestDelay.Value > base.RateLimit.TotalSeconds)
+                {
+                    return TimeSpan.FromSeconds(definition.RequestDelay.Value);
+                }
+
+                return base.RateLimit;
+            }
+        }
 
         public override IIndexerRequestGenerator GetRequestGenerator()
         {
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Indexers/Definitions/Cardigann/Cardigann.cs
```diff
@@ -33,6 +33,21 @@ public class Cardigann : TorrentIndexerBase<CardigannSettings>
         // thinking its a partial page and insteaad all search_path requests are run for each indexer
         public override int PageSize => 1;
 
+        public override TimeSpan RateLimit
+        {
+            get
+            {
+                var definition = _definitionService.GetCachedDefinition(Settings.DefinitionFile);
+
+                if (definition.RequestDelay != null && definition.RequestDelay.Value > base.RateLimit.TotalSeconds)
```
Qstick, line 42: Lets use `.HasValue` instead of the null comparison on nullable property

Qstick (changes requested):

Qstick:
This looks fine, just the one comment for consistency. 

On a separate but somewhat similar note, we don't currently rate limit auth or download requests for any indexer. I've debated the auth side due to request speed being affected when we have to auth mid request. The download we should likely do (I want to do a bit of a rewrite to the download code in the base to make this easier if we decide on it). 

Thoughts on these two?

Qstick (approved):

Qstick:
Yea, The rate limit is controlled at the request level, and not the client level for Prowlarr. If you want to give it a test, breakpoint here:

https://github.com/Prowlarr/Prowlarr/blob/71dd8b6d04eb477181d9e83ff2412a2cff9d05f8/src/NzbDrone.Common/Http/HttpClient.cs#L161-L164

In my testing, I can confirm that most auth requests and download requests are not passed with a RateLimit (set at the request level) thus they shoot right though. 

image

image

Qstick:
Well the part I find odd is that the only way httpClient picks it up is if it’s set in the request, and the only place we set it in the request is in the search method here

https://github.com/Prowlarr/Prowlarr/blob/364a5564ae169f64fc55a41a40847fbdb5724c01/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs#L342

So really the download should always be just the hard-coded 2 seconds from download service.

PR #1291, opened 2023-01-05
Fixed: (GreatPosterWall) Use UTC for PublishDate, order releases

mynameisbogdan:
and map categories accordingly

#### Database Migration
NO

#### Description
On Windows `DateTimeOffset().LocalDateTime` seems to be okay, but on Linux it seems to think my localtime it's UTC then adds the difference once again, thus doubling my timezone. So let's try `DateTimeOffset().UtcDateTime`.

Also category mapping by resolution, just like Avistaz does.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/GreatPosterWall.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/GreatPosterWall.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using Newtonsoft.Json;
 using NLog;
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Definitions/GreatPosterWall.cs
```diff
@@ -186,6 +191,20 @@ protected string GetDownloadUrl(int torrentId, bool canUseToken)
...
         return url.FullUri;
     }
+
+    protected virtual List<IndexerCategory> ParseCategories(GreatPosterWallTorrent torrent)
+    {
+        var cats = new List<IndexerCategory>();
+
+        cats.Add(torrent.Resolution switch
+        {
+            var res when _hdResolutions.Contains(res) => NewznabStandardCategory.MoviesHD,
```
Qstick, line 201: Are we able to add similar category capabilities to searching (like Avistaz). If we are returning categories for these results it would be have these categories in the indexers capabilities categories list and be able to search with them.

Qstick (approved):

PR #1293, opened 2023-01-05
Fixed: (Torznab) Use HashSet to prevent duplicated flags

mynameisbogdan:
#### Database Migration
NO

Before:
image

After:
image


Qstick (commented):
Looks fine, but wouldn't hurt to change this on the `ReleaseInfo` base class instead of just this one instance to avoid this happening in other indexers

Qstick:
Cool thanks, that should be fine. Will merge after build.

Qstick (approved):

Issue #1257, opened 2022-12-25
Show Ratio 


techguruz:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Would be nice if you could see your ratio and a few other stat maybe user group currently seeding/leeching.

### Describe the solution you'd like

It should be pretty easy to scrape at the very least your ratio but since most trackers use the same framework ratio leeching/seeding should be able to be pulled too maybe even more stuff. Once it has been scraped it could be saved to a database then displayed on the stats page to you can see your upload/download amount and ratio change since the last time you viewed or maybe weekly/monthly scrape so your not constantly hitting the trackers up for that info. 

### Describe alternatives you've considered

I have previously used pt-plugin-pro.

### Anything else?

https://github.com/pt-plugins/PT-Plugin-Plus

[AB#4547](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/4547)

Qstick:
Covered by #1203

PR #1299, opened 2023-01-07
Fixed: (MyAnonamouse) Added search type options, search in narrator/series as settings

### LINKED ISSUE #810: Add "Search for freeleech only" to MAM

Just want to browse freeleech on prowlarr.

Add option to search for freeleech only on Myanonamouse. This would be the same check box used on Torrentleech and others.

Search page could alternatively have a checkbox for freeleech, or Tracker flags in general

### Anything else?

I can see freeleech flags on the search results page, but no way to filter on that, or to specify before searching.

[AB#2516](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/2516)

mynameisbogdan:
#### Database Migration
NO

#### Description
- `ExcludeVip` was a little confusing for me, so I replaced it with `SearchType` which contains multiple options to choose from.
Screenshot 2023-01-07 at 19 51 52
- fixed pagination since their API supports it.
- add search in narrator by default, add description and series as settings
- rename title of the setting `Freeleech` to something more descriptive
- show first 5 authors in title, not just the first one
- use description
- removed double slash from links

So far I'm still happy with `DownloadVolumeFactor = item.Free ? 0 : 1`, which should be something like `DownloadVolumeFactor = item.Free ? 0 : item.FreeVip ? 0 : 1` if you're VIP.
I don't know if to use a boolean setting to check `FreeVip` or a date, but since you can buy VIP all the time with bonus points it's a little annoying to always change the date since it's limited to a maximum of days (90 I think). 
Another option would be to do another request first to `/jsonLoad.php` to check for user class if it's VIP...
Would love to hear suggestions on this regard.

Fixes #810

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/MyAnonamouse.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/MyAnonamouse.cs
@@ -9,6 +9,8 @@ using System.Threading.Tasks;
 using FluentValidation;
 using Newtonsoft.Json;
 using NLog;
+using NzbDrone.Common.Cache;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 using NzbDrone.Common.Serializer;
 using NzbDrone.Core.Annotations;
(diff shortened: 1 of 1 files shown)


Qstick:
> I don't know if to use a boolean setting to check `FreeVip` or a date, but since you can buy VIP all the time with bonus points it's a little annoying to always change the date since it's limited to a maximum of days (90 I think).
> Another option would be to do another request first to `/jsonLoad.php` to check for user class if it's VIP...
> Would love to hear suggestions on this regard.

Yea, agree, setting the date is a bit of a burden. It would be nice to do this in an automated way. Along the lines of #1257 it would be cool at some point if Prowlarr was aware of user status and ratio and could make decisions based on that, for now I think you'll have to make the separate request. If the extra requests are a problem for user limits we can cache it for a day or more to avoid doing it on every search (similar to what we do with newznab capabilities).

Qstick:
As an additional note, There's also a VipExpiration healthcheck that will look for a setting titled `VipExpiration` on the indexer. If we were to store this in a hidden settings after we grab it, that logic should pick it up and warn the user of expiration 1 week before hand. Eventually it would be nice to auto-grab expiration from newznab indexers that have this setting similar to what you are doing here.

src/NzbDrone.Core/Indexers/Definitions/MyAnonamouse.cs
```diff
@@ -210,84 +213,96 @@ public class MyAnonamouseRequestGenerator : IIndexerRequestGenerator
...
+            {
+                parameters.Add("tor[srchIn][narrator]", "true");
+            }
+
+            if (Settings.SearchInSeries)
+            {
+                parameters.Add("tor[srchIn][series]", "true");
+            }
+
+            if (Settings.SearchInFilenames)
```
Qstick, line 254: Is there any case for always searching in all 3 of these instead of having it user selectable? Just thinking about potential support cases because of varying results. I'm fine leaving it as setting if we think it has value.

Qstick (approved):

src/NzbDrone.Core/Indexers/Definitions/MyAnonamouse.cs
Qstick: Nah, we can leave those. Just trying to think about best reliability out of the box. In many cases users won't go look at the settings, or they start toggling them not knowing what they do then come for support. 

Seems like your fixups are best of both worlds here.

Qstick (approved):

PR #1309, opened 2023-01-09
Fixed: (AvistaZ) Use different timezone offset than the rest

mynameisbogdan:
#### Database Migration
NO

#### Description
AvistaZ was off by a large margin, so let's use the correct timezone offset for it. And in the future `TimezoneOffset` can be overwritten per tracker if needed.

Also some CS fixes, which I'll undo if they aren't necessary. 👍

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/AvistaZ.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/AvistaZ.cs
@@ -1,6 +1,5 @@
 using System.Collections.Generic;
 using NLog;
-using NzbDrone.Common.Http;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.Indexers.Definitions.Avistaz;
 using NzbDrone.Core.Messaging.Events;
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazBase.cs
```diff
@@ -13,12 +13,12 @@ namespace NzbDrone.Core.Indexers.Definitions.Avistaz
     public abstract class AvistazBase : TorrentIndexerBase<AvistazSettings>
     {
         public override DownloadProtocol Protocol => DownloadProtocol.Torrent;
-        public override string[] IndexerUrls => new string[] { "" };
-        protected virtual string LoginUrl => Settings.BaseUrl + "api/v1/jackett/auth";
+        public override string[] IndexerUrls => new[] { "" };
```
Qstick, line 16: Can we drop this override from the base here to force the need to create it on all derived classes.

src/NzbDrone.Core/Indexers/Definitions/Avistaz/AvistazParser.cs
```diff
@@ -61,7 +58,7 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                     InfoUrl = details,
                     Guid = details,
                     Categories = cats,
-                    PublishDate = DateTime.Parse(row.CreatedAt + "-05:00").ToUniversalTime(), // Avistaz does not specify a timezone & returns server time
+                    PublishDate = DateTime.Parse($"{row.CreatedAt} {TimezoneOffset}").ToUniversalTime(),
```
Qstick, line 61: Should we convert to use `DateTimeStyles.AdjustToUniversal` in the parse. Docs seem to indicate it's a bit more efficient


> The same effect can be achieved by calling the [DateTime.ToUniversalTime](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.touniversaltime?view=net-7.0) or [DateTimeOffset.ToUniversalTime](https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.touniversaltime?view=net-7.0) method. However, using the [AdjustToUniversal](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.datetimestyles?view=net-7.0#system-globalization-datetimestyles-adjusttouniversal) value with the date and time parsing method is more efficient.

Qstick (changes requested):
Nice cleanup here, looks good. Just a few comments

src/NzbDrone.Core/Indexers/Definitions/AvistaZ.cs
```diff
@@ -10,18 +9,24 @@ namespace NzbDrone.Core.Indexers.Definitions
...
     {
         public override string Name => "AvistaZ";
-        public override string[] IndexerUrls => new string[] { "https://avistaz.to/" };
+        public override string[] IndexerUrls => new[] { "https://avistaz.to/" };
         public override string Description => "Aka AsiaTorrents";
         public override IndexerPrivacy Privacy => IndexerPrivacy.Private;
 
-        public AvistaZ(IIndexerRepository indexerRepository, IIndexerHttpClient httpClient, IEventAggregator eventAggregator, IIndexerStatusService indexerStatusService, IConfigService configService, Logger logger)
+        public AvistaZ(
+            IIndexerRepository indexerRepository,
```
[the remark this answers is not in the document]
Qstick: Yep, your example is spot on, that's in line with what we've been trying to do in code base on all 4 apps. Probably better to follow StyleCop in time, it just makes it more difficult to cherry-pick between the 4 when we start making small tweaks to one. (though in this case that doesn't really apply)

Let's stick with it here, then we can make a change to turn on SA1116 in editorconfig in a separate commit and convert the code base.

Qstick (approved):

PR #1318, opened 2023-01-11
Fixed: (UI) Transpile chart.js to fix blank page on some browsers

mynameisbogdan:
#### Database Migration
NO

#### Description
chart.js v4 supports only modern browsers and so per [their recommendation](https://github.com/chartjs/Chart.js/issues/11028#issuecomment-1378391109) I'm allowing babel to transpile their library.

I'm getting `SyntaxError: Unexpected token '='. Expected an opening '(' before a method's parameter list.` due to code like https://github.com/chartjs/Chart.js/blob/b2b881b9b104c4ce0699c64d8037c4b7fa7cae3e/src/core/core.datasetController.js#L224-L229

Related: chartjs/Chart.js#11028

Fixing the regex for extensions as well since I don't think you're planning to support `.j` files.

### DIFF
--- a/frontend/build/webpack.config.js
+++ b/frontend/build/webpack.config.js
@@ -142,8 +142,8 @@ module.exports = (env) => {
     module: {
       rules: [
         {
-          test: /\.js?$/,
-          exclude: /(node_modules|JsLibraries)/,
+          test: /\.jsx?$/,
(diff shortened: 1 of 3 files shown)


Qstick (approved):
But I was planning to re-write the UI with `.j` :D

Thanks

Qstick:
At what point do you think we stop worrying about ES6 compat?

PR #1324, opened 2023-01-14
Fixed: (SpeedCD) Add wildcard to season in tvsearch, add freeleech toggle, improve query selectors

mynameisbogdan:
Database Migration
YES - 25

#### Description
- `tvsearch` was searching only by `imdbid` returning all seasons and episodes available, and changed to if you search season `S02` it will lookup for `imdbid S02*` or `tvshow S02*`, `*` being wildcard. 
- Same for episode, thus if you search an episode it will return only that episode.
- I should mention that I left the wildcard for episodes too, since the worst that could happen is to return E10X for E10 and so on.

Hope it makes sense.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/SpeedCD.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/SpeedCD.cs
@@ -7,7 +7,6 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using AngleSharp.Html.Parser;
-using FluentValidation;
 using NLog;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Indexers/Definitions/SpeedCD.cs
```diff
@@ -19,14 +18,13 @@
 using NzbDrone.Core.Messaging.Events;
 using NzbDrone.Core.Parser;
 using NzbDrone.Core.Parser.Model;
-using NzbDrone.Core.Validation;
 
 namespace NzbDrone.Core.Indexers.Definitions
 {
-    public class SpeedCD : TorrentIndexerBase<UserPassTorrentBaseSettings>
+    public class SpeedCD : TorrentIndexerBase<SpeedCDSettings>
```
Qstick, line 24: This will need a DB migration since it's changing the ConfigContract for the Indexer. All else looks good here.

Qstick: Yep, that works

Qstick (approved):

PR #1327, opened 2023-01-14
Fixed: (Rarbg) Add slash to IndexerUrl, increase RateLimit to 5s

mynameisbogdan:
#### Database Migration
NO

#### Description
I don't think it's in fact 1req/2s like they say in `apidocs_v2.txt`, I think they also use something like `token bucket algorithm` with a span of a few minutes.

- ~~Let's increase the RateLimit to 5s and to retry when `rate_limit=1` is in the error response.~~
- log why the parser didn't find any results.

/cc @bakerboy448

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Rarbg/RarbgParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Rarbg/RarbgParser.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Text.RegularExpressions;
+using NLog;
 using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Common.Http;
 using NzbDrone.Core.Indexers.Exceptions;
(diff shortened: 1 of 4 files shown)


Qstick:
The retry on `rate_limit=1` is in fact what we had prior to this <https://github.com/Prowlarr/Prowlarr/commit/518c85dee2bf05ccc840c1e2f87ff696f77672af>. Which attempted to make the changes that rarbg admins wanted with our rate limit. The issue is they have some bugs in when they use `rate_limit=1` in some cases, as it's returned on normal 200 responses with results. So part of that was reversed here 5cc044aa8fbc7ba62d9c32442d8125b9b347d28a. The rate_limit=1 retry was not put back though, It's like fine to add this back, but we may want to check to see if it's an actual good response with `rate_limit=1` before we retry.

src/NzbDrone.Core/Indexers/Definitions/Rarbg/RarbgParser.cs
```diff
@@ -41,12 +44,17 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
...
-                    // No results, rate limit, or imdbid/tvdb not found
+                    if (jsonResponse.Resource.rate_limit is 1)
+                    {
+                        _logger.Debug("No results due to rate limiting.");
+                    }
+                    else
+                    {
+                        _logger.Debug("No results or imdbid/tvdb not found: error code {0}.", jsonResponse.Resource.error_code);
+                    }
+
```
Qstick, line 57: Would it be better to log out the text error they provide as well `jsonResponse.Resource.error`

Qstick (approved):

PR #1337, opened 2023-01-18
Fixed: (TorrentDay) Add freeleech only setting

### LINKED ISSUE #1335: TorrentDay Freeleech checkbox

Some of the other indexers such as TorrentLeech have a "Search freeleech only" checkbox.
When searching in Prowlarr using indexer TorrentDay it recognizes a freeleech flag, but there is no way to turn this on.

It would be helpful if I could have a freeleech option, so that I can run 2 versions of it, with different priority amounts to use with Sonarr. Not really a issue with Radarr since you can search with flags.
Thank you.

Couldn't think of anything other than searching manually, since it seems Sonarr is currently not supporting this.

### Anything else?

I honestly dont know how else to solve this. It might not even solve it, since I think priorities and scores work very differently. But since other indexers have this option, figured it could be easily implemented for this also.
Thank you

[AB#4751](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/4751)

mynameisbogdan:
#### Database Migration
YES - 25

#### Description
Adds freeleech only setting.

I don't have an account to verify, but according to user's screenshot it only needs `;free` added for FL only search even if Jackett has `;free=on`. 

https://github.com/Prowlarr/Prowlarr/issues/1335#issuecomment-1386381234

* Fixes #1335

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/026_torrentday_cookiesettings_to_torrentdaysettings.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/026_torrentday_cookiesettings_to_torrentdaysettings.cs
@@ -1,0 +1,14 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(026)]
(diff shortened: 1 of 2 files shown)


Qstick (approved):
`;free` confirmed to work

PR #1341, opened 2023-01-18
New: Add apps minimum seeders per indexer

mynameisbogdan:
#### Database Migration
NO

#### Description
Sometimes the app's minimum seeders value isn't one size fits all for every indexer, so I think this would be a good addition to be able to set it per indexer if you'll like, with fallback to Sync profile's value.

### DIFF
--- a/src/NzbDrone.Core/Indexers/IndexerTorrentBaseSettings.cs
+++ b/src/NzbDrone.Core/Indexers/IndexerTorrentBaseSettings.cs
@@ -8,6 +8,10 @@ namespace NzbDrone.Core.Indexers
     {
         public IndexerTorrentSettingsValidator(double seedRatioMinimum = 0.0, int seedTimeMinimum = 0, int seasonPackSeedTimeMinimum = 0)
         {
+            RuleFor(c => c.AppMinimumSeeders).GreaterThan(0)
+                .When(c => c.AppMinimumSeeders.HasValue)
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Indexers/IndexerTorrentBaseSettings.cs
```diff
@@ -38,13 +42,16 @@ public class IndexerTorrentBaseSettings
     {
         private static readonly IndexerTorrentSettingsValidator Validator = new IndexerTorrentSettingsValidator();
 
-        [FieldDefinition(1, Type = FieldType.Textbox, Label = "Seed Ratio", HelpText = "The ratio a torrent should reach before stopping, empty is app's default", Advanced = true)]
+        [FieldDefinition(1, Type = FieldType.Number, Label = "Minimum Seeders", HelpText = "Minimum seeders required for the indexer to grab, empty is app's default", Advanced = true)]
```
bakerboy448, line 45: *Sync profile's default

Qstick:
One reason I've been a bit hesitant on this is that it implies that it will apply to all grabs for an indexer, when in reality it only applies to those that come from the apps (i.e. this has no effect on anything outside of app syncing). We should likely clarify that somehow.

bakerboy448:
`Synced App Minimum Seeders` is probably a bit wordy :/

Qstick (approved):

PR #1345, opened 2023-01-20
New: (Notification) Apprise

mynameisbogdan:
#### Database Migration
NO

#### Description
Notification integration with [Apprise](https://github.com/caronc/apprise).


Qstick:
This looks good! Nice addition. Is there anything we can do to retrieve acceptable values for Tags and StatelessUrls to make this a bit more user-friendly, or is that difficult to do from their API

Qstick:
> 

Manual is fine if we think that's the way to go. I'm unfamiliar with the API and setup, just throwing out the idea if it will help avoid user setup issues that we've seen with other providers.

Qstick:
If it makes more sense to leave tags as manual entry then I'm fine with it

Qstick (approved):

PR #1347, opened 2023-01-21
New: Add Toloka.to

mynameisbogdan:
#### Database Migration
NO

#### Description
It's based on Jackett's version, but with a little more title cleanup.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
@@ -1,0 +1,469 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using AngleSharp.Html.Parser;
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
```diff
@@ -0,0 +1,469 @@
...
+            return pageableRequests;
+        }
+
+        public IndexerPageableRequestChain GetSearchRequests(TvSearchCriteria searchCriteria)
+        {
+            var pageableRequests = new IndexerPageableRequestChain();
+
+            var term = $"{searchCriteria.SanitizedSearchTerm}";
+
+            if (_capabilities.TvSearchSeasonAvailable && searchCriteria.Season is > 0)
```
Qstick, line 294: Won't `_capabilities.TvSearchSeasonAvailable` always be true? It's just looking at the capabilities that are setup above.

Qstick (approved):
Looks good, just the one comment on season logic.

Qstick (approved):

PR #1351, opened 2023-01-22
New: Add Sharewood API

mynameisbogdan:
#### Database Migration
NO

#### Description
The C# implementation for sharewood, sadly with limitations. For example their API doesn't support freeleech only or filtering by multiple categories.

Default limit is 25 for pages with category id, 100 for without category id.

The class name is `Sharewood`, should I renamed it to `SharewoodAPI` or setting the name to `Sharewood API` is enough?

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Sharewood.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Sharewood.cs
@@ -1,0 +1,442 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Globalization;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Text.RegularExpressions;
+using FluentValidation;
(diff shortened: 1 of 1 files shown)


Qstick:
Is there any benefit in this if the YML already has more functionality? Also if there is, can this be done in YML?

Qstick:
> You can read the conversation they had with Jackett [here](https://github.com/Jackett/Jackett/issues/10269).

Got it, Thanks! So the regex is the main reason this needs to be C# instead of using YML for the API version as well it sounds like?

Qstick:
> No, the most important reason is because their staff is requesting use of their API, and not to scrape. As you can see in that ticket. 
> 
> We can postpone this PR until their API is able to filter by multiple categories in one request.

We can use the API via a yml, my question was more of can we use yml for this API version too, or does it HAVE to be c#?

PR #1365, opened 2023-01-27
Fixed: (LazyLibrarian) Error converting null to bool

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix for `[v10.0.0.8245] Newtonsoft.Json.JsonSerializationException: Error converting value {null} to type 'System.Boolean'. Path 'Data.Newznabs[0].ENABLED', line 1, position 91.`


Qstick:
Do we need to null check it before we send it back as a string here

```
            var parameters = new Dictionary<string, string>
            {
                { "name", indexer.Name },
                { "providertype", indexer.Type.ToString().ToLower() },
                { "host", indexer.Host },
                { "prov_apikey", indexer.Apikey },
                { "enabled", indexer.Enabled.ToString().ToLower() },
                { "categories", indexer.Categories }
            };
```

PR #1375, opened 2023-01-31
Fixed: (GreatPosterWall) Parse categories based on resolution

mynameisbogdan:
#### Database Migration
NO


Qstick (commented):
Do these need to be add to capabilities?

Qstick (approved):

PR #1384, opened 2023-02-02
Fixed: (CloudFlareDetection) Check for DDoS-Guard case-insensitive

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix for `<title>DDoS-Guard</title>`.


Qstick:
Have you seen this in the wild? If so FlaresolverrSharp (the library jackett uses for similar code) also needs an update

Qstick (approved):

PR #1385, opened 2023-02-02
Fixed: (GreatPosterWall) Remove special characters from titles

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix for
image

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/GreatPosterWall.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/GreatPosterWall.cs
@@ -239,180 +239,76 @@ public class GreatPosterWallSettings : GazelleSettings
 
 public class GreatPosterWallResponse
 {
-    [JsonProperty("status")]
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Definitions/GreatPosterWall.cs
```diff
@@ -248,171 +248,171 @@ public class GreatPosterWallResponse
...
-        public List<Result> Results { get; set; }
+    [JsonProperty("results")]
+    public List<Result> Results { get; set; }
 }
 
 public class Result
 {
-        [JsonProperty("groupId")]
-        public int GroupId { get; set; }
+    [JsonProperty("groupId")]
```
Qstick, line 263: Really we can probably get rid of the `JsonProperty` attributes on all of these. They should work by default given they are just CamelCase

Qstick (approved):

Issue #1277, opened 2023-01-02
ToDo: Restore Rarbg HTTP 200 Rate Limiting [Waiting on Rarbg]


bakerboy448:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Restore HTTP 200 & rate_limit 1  as too many requests.

reverted https://github.com/Prowlarr/Prowlarr/commit/5cc044aa8fbc7ba62d9c32442d8125b9b347d28a#diff-bd56c63a2154391186c3916be78aa2c0b6de3dca774930b0a4cb99b29d1b8425L46  

Due to false positive rate_limit of 1 with 200s when no rate limit was actually occurring (results were found)

awaiting internal fix from rarbg for false positive 

### Describe the solution you'd like

`if the API returns 200 with a json that contains rate_limit 1 retry in 5 minutes`
https://github.com/Prowlarr/Prowlarr/issues/1169

### Describe alternatives you've considered

N/A

### Anything else?

N/A

bakerboy448:
for posterity and context in the future

              there are multiple levels of late limiters
cloudflare -> web -> code

when rated limited cloudflare will return http code 429 with a json encoded string that includes rate limit 1
when rated limited web returns http code 444 to cloudflare which returns 520 on cloudflare , there is no way to send 429 to cloudflare without it breaking the request properly and retrying it
when rated limited code returns http code 200 with rate_limit 1

rate_limit : 1 is generally returned when you've exceeded the rate limit or when one of the many servers behind cloudflare is being overloaded - because cloudflare doesn't evenly redistribute requests equally its there so clients via api can see that they are rate limited , retry in lets say 5 seconds and they can get their request fullfilled on another backend ( cloudflare doesn't do stickiness ) in x time

_Originally posted by rarbg in https://github.com/Jackett/Jackett/issues/13663#issuecomment-1288190301_

PR #1410, opened 2023-02-09
New: (RarBg) Log Rate Limiting at Info

bakerboy448:
#### Database Migration
NO

#### Description

Most (new) prowlarr users using rarbg are likely not running debug....and this causes extra support across all starrs.   Log no results details at info so it's clear even to users 
#### Todos
- Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Rarbg/RarbgParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Rarbg/RarbgParser.cs
@@ -50,11 +50,11 @@ namespace NzbDrone.Core.Indexers.Rarbg
 
                     if (jsonResponse.Resource.rate_limit is 1)
                     {
-                        _logger.Debug("No results due to rate limiting. Reason: {0}", reason);
+                        _logger.Info("No results due to rate limiting. Reason: {0}", reason);
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


Qstick:
I can see the rate limit being info, but the other seems spammy

src/NzbDrone.Core/Indexers/Definitions/Rarbg/RarbgParser.cs
```diff
@@ -50,7 +50,7 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
 
                     if (jsonResponse.Resource.rate_limit is 1)
                     {
-                        _logger.Debug("No results due to rate limiting. Reason: {0}", reason);
+                        _logger.Info("No results due to rate limiting. Reason: {0}", reason);
```
bakerboy448, line 53: Actually should this be a warning instead?

Qstick: As much as they rate limit, no, I don't think so

bakerboy448: should we be throwing ratelimitreached or is https://github.com/Prowlarr/Prowlarr/issues/1277 still occurring?

PR #1431, opened 2023-02-15
Fixed: (SpeedAppBase) Add pagination

mynameisbogdan:
#### Database Migration
NO

#### Description
This indexer returns so many irrelevant results and still tries to fetch 10 pages of page=1, let's try to "fix" it with pagination to fetch fewer pages.

Btw is there a way to tell Sonarr to fetch for example maximum 500 results, not 1000?


bakerboy448:
Sonarr logic per Markus

> If there are more than a certain number of results 90/100? Sonarr tries to get the next page, if the indexer keeps returning crap it'll keep trying up to the maximum number of pages, which is 30 it looks like

Qstick (approved):

PR #1439, opened 2023-02-17
Fixed: (TorrentIndexerBase) Validate downloaded torrent data

mynameisbogdan:
#### Database Migration
NO

#### Description
I'm having issues with NBL on my "prod" installation returning `HTTP/1.1 200 OK`&`Content-Length: 0`, the logs don't show anything and I can't replicate it on my dev machine. :)


src/NzbDrone.Core/Indexers/Definitions/Cardigann/Cardigann.cs
```diff
@@ -233,6 +233,7 @@ public override async Task<byte[]> Download(Uri link)
                 throw;
             }
 
+            ValidateTorrent(downloadBytes);
```
Qstick, line 236: Can we add new line before return on these.

Qstick (changes requested):

Qstick (approved):

Issue #536, opened 2021-10-06
Indexer - ShowRSS - Returning Unrelated Results


bakerboy448:
Reference the ToDo in code:

https://github.com/Prowlarr/Prowlarr/blob/1bfcb99f313d86dabda48b9799306c745943efd1/src/NzbDrone.Core/Indexers/Definitions/ShowRSS.cs#L144

X-Ref https://github.com/Prowlarr/Prowlarr/issues/535

[AB#1743](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/1743)

slimjim91:
Looks like a similar problem to what's happening with EZTV. Searching anything brings up unrelated junk. Seems like it needs the same post-filter

bakerboy448:
@slimjim91 

Not at all

ShowRSS is C# and EZTV is not. EZTV is Cardigann/YML.

They have no relation to each other.

See https://github.com/Prowlarr/Indexers/pull/77
https://github.com/Prowlarr/Indexers/issues/78

Qstick:
It's an RSS feed, it's not meant to be used for searches.

bakerboy448:
@mynameisbogdan thoughts on purging ShowRSS and adding it as a default for generic RSS?

PR #1450, opened 2023-02-19
Filter search criteria

Qstick:
#### Database Migration
NO

#### Description
Fix IsRss to work properly with new query options
Add method to filter search results by query if is not an id search

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/Indexers/IndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/IndexerBase.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using FluentValidation.Results;
 using NLog;
@@ -104,7 +105,7 @@ namespace NzbDrone.Core.Indexers
 
         public abstract IndexerCapabilities GetCapabilities();
 
-        protected virtual IList<ReleaseInfo> CleanupReleases(IEnumerable<ReleaseInfo> releases)
+        protected virtual IList<ReleaseInfo> CleanupReleases(IEnumerable<ReleaseInfo> releases, SearchCriteriaBase searchCriteria)
         {
             var result = releases.ToList();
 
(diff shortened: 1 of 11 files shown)


src/NzbDrone.Core/Indexers/IndexerBase.cs
```diff
@@ -146,6 +147,24 @@ protected virtual IList<ReleaseInfo> CleanupReleases(IEnumerable<ReleaseInfo> re
...
+
+            if (!searchCriteria.IsIdSearch)
+            {
+                var splitRegex = new Regex("[^a-zA-Z0-9]+");
+
+                // split search term to individual terms for less aggressive filtering, filter common terms
+                var terms = splitRegex.Split(searchCriteria.SearchTerm).Where(t => !commonWords.Contains(t) && t.IsNotNullOrWhiteSpace() && t.Length > 1);
+
+                // check in title and description for any term searched for
+                releases = releases.Where(r => terms.Any(t => r.Title.Contains(t) || r.Description.Contains(t))).ToList();
```
Qstick, line 162: Should we look at additional release properties for the terms. Ex. user searches by imdb id but doesn't use the imdb field. By comparing term to `release.imdb` or `release.tmdb` we can possible catch those from being filtered out.

Qstick: Right agree, however this is an `or` logic, so if anything matches it will accept the release (title, description, etc). Adding the ids would just give one more possible match criteria, and not necessarily cause a match to be removed. Meaning fewer valid releases may get filtered (less false positives).

src/NzbDrone.Core/Indexers/IndexerBase.cs
```diff
@@ -146,6 +147,24 @@ protected virtual IList<ReleaseInfo> CleanupReleases(IEnumerable<ReleaseInfo> re
             return result.DistinctBy(v => v.Guid).ToList();
         }
 
+        protected IEnumerable<ReleaseInfo> FilterReleasesByQuery(IEnumerable<ReleaseInfo> releases, SearchCriteriaBase searchCriteria)
+        {
+            var commonWords = new[] { "a", "an", "the" };
+
+            if (!searchCriteria.IsIdSearch)
+            {
+                var splitRegex = new Regex("[^a-zA-Z0-9]+");
```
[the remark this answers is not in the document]
Qstick: Fixed, good call

bakerboy448:
Fixes https://github.com/Prowlarr/Prowlarr/issues/536 ?

Qstick:
> Fixes #536 ?

Yea, It can if we add it on ShowRSS. I didn't too a full global (though it would be easy if we feel like the method is good enough to avoid false positives).

PR #1464, opened 2023-02-22
Fixed: (Cardigann) Parse text templates only if necessary

mynameisbogdan:
#### Database Migration
NO

Removed a try-catch, so use `?w=1` to diff this PR.


src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannBase.cs
```diff
@@ -349,11 +349,13 @@ protected string HandleJsonSelector(SelectorBlock selector, JToken parentObj, Di
 
         public string ApplyGoTemplateText(string template, Dictionary<string, object> variables = null, TemplateTextModifier modifier = null)
         {
-            if (variables == null)
+            if (string.IsNullOrWhiteSpace(template) || !template.Contains("{{"))
```
Qstick, line 352:
```suggestion
            if (template.IsNullOrWhiteSpace() || !template.Contains("{{"))
```

Qstick (approved):

PR #1469, opened 2023-02-23
Fixed: (Cardigann) Switch to DateTime standard

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannBase.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannBase.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannBase.cs
```diff
@@ -613,14 +614,22 @@ protected string ApplyFilters(string data, List<FilterBlock> filters, Dictionary
                     case "timeparse":
                     case "dateparse":
                         var layout = (string)filter.Args;
-                        try
+
+                        if (layout.Contains("yy") && DateTime.TryParseExact(data, layout, CultureInfo.InvariantCulture, DateTimeStyles.None, out var parsedDate))
```
Qstick, line 618: Can we just remove the replaces in parseGoLang and enforce the correct format in definitions with yml?

Qstick (approved):

PR #1478, opened 2023-02-25
Fixed: (Cardigann) Prevent fetching the first page multiple times

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannRequestGenerator.cs
@@ -59,7 +59,7 @@ namespace NzbDrone.Core.Indexers.Cardigann
             variables[".Query.TraktID"] = searchCriteria.TraktId?.ToString() ?? null;
             variables[".Query.DoubanID"] = searchCriteria.DoubanId?.ToString() ?? null;
 
-            pageableRequests.Add(GetRequest(variables));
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannParser.cs
```diff
@@ -373,7 +373,7 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
 
             _logger.Debug($"Got {releases.Count} releases");
 
-            return releases;
+            return releases.OrderByDescending(o => o.PublishDate).ToArray();
```
Qstick, line 376: Some indexers have sort settings, does this make it confusing when results are not in same order?

src/NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannRequestGenerator.cs
```diff
@@ -1015,8 +1015,14 @@ public bool CheckIfLoginIsNeeded(HttpResponse response)
             return false;
         }
 
-        private IEnumerable<IndexerRequest> GetRequest(Dictionary<string, object> variables)
+        private IEnumerable<IndexerRequest> GetRequest(Dictionary<string, object> variables, int limit, int offset)
```
Qstick, line 1018: just pass in `searchCriteria` here, then we only need to default the offset and limit once instead of repeating it in every method call above.

Qstick (approved):

PR #1482, opened 2023-02-26
Fixed: Disable HealthChecks when in testing runtime

mynameisbogdan:
#### Database Migration
NO

#### Description
Disabled system time and notification health checks while running in testing runtime.


Qstick (approved):

Qstick:
`BuildInfo.IsDebug` might be better here. 

Otherwise you really want `!RuntimeInfo.IsProduction`, but these are static methods, so you won't be able to mock them to fix the system time test which fails given it expects an error, but never gets far enough in the method now to get one.

Qstick (approved):

Issue #500, opened 2021-09-13
[FEAT]: Sync indexer tags to indexers in apps || Add Application Tags to Indexers


6cUbi57z:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

I currently use indexer tags to filter the indexers used for various series. While I can add these in Prowlerr, they don't seem to sync to the apps.

### Describe the solution you'd like

Tags applied in Prowlerr are automatically synced to apps.

### Describe alternatives you've considered

Adding tags manually removes them on next sync.

### Anything else?

Seems odd for this field to be mandatory...

[AB#1579](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/1579)

**EDIT:** Corrected info on alternatives attempted. Forgot it was in "Add and remove only" sync mode so hadn't updated.

santhorus:
This would be really usefull, it is getting implemented?

FabrizioMFE:
Is there any plans on implementing this in the near future? The issue basically defeats the purpose of using full sync

6cUbi57z:
Forgot I had started looking at this. Just got an implementation working for Sonarr and Radarr. I don't have instances available for the other apps.

datcactus:
Seems like a useful feature, but based on @bakerboy448 's comment in #1332 , I can't help but wonder if it's a better idea to preserve the existing tags on the app side vs syncing them over, as nice as a full sync might be.

FabrizioMFE:
Well from what I see in that PR, it would be a toggle on the Apps settings, maybe on top of the toggle it could also be that Prowlarr just doesn't override tags unless that toggle is on?

When option on => Sync tags from Prowlarr indexers to Apps indexers
When option off => Prowlarr doesn't touch tags in Apps indexers at all

6cUbi57z:
Hmm... Good point. At the moment sync being off would default to overwriting indexers to have no tags. If the setting is off it should definitely leave them alone. That's a bug. 

I'd be hesitant to have tag sync as append only. The full sync option in prowlarr is supposed to keep everything up to date AFAIK and having the setting only append then risks other future settings doing the same and "full sync" becomes far less effective. I would think if append only is wanted, the app would be set to add and remove only rather than full sync.

averon420:
[...]

PR #1490, opened 2023-02-27
Fixed: (Apps) Don't clear user defined tags

mynameisbogdan:
#### Database Migration
NO

#### Description
- Keep user defined tags on remote indexer on sync
- Fixed the log message for `AddIndexer` showing `Skipping add for indexer` when it was actually successful.
- Improved the debug logs a little bit and some CS fixes.

### DIFF
--- a/src/NzbDrone.Core/Applications/Whisparr/Whisparr.cs
+++ b/src/NzbDrone.Core/Applications/Whisparr/Whisparr.cs
@@ -61,7 +61,7 @@ namespace NzbDrone.Core.Applications.Whisparr
         public override List<AppIndexerMap> GetIndexerMappings()
         {
             var indexers = _whisparrV3Proxy.GetIndexers(Settings)
-                                         .Where(i => i.Implementation == "Newznab" || i.Implementation == "Torznab");
+                .Where(i => i.Implementation == "Newznab" || i.Implementation == "Torznab");
 
             var mappings = new List<AppIndexerMap>();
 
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Core/Applications/Whisparr/Whisparr.cs
```diff
@@ -84,15 +84,19 @@ public override List<AppIndexerMap> GetIndexerMappings()
 
         public override void AddIndexer(IndexerDefinition indexer)
         {
-            if (indexer.Capabilities.Categories.SupportedCategories(Settings.SyncCategories.ToArray()).Any())
+            if (!indexer.Capabilities.Categories.SupportedCategories(Settings.SyncCategories.ToArray()).Any())
```
Qstick, line 87: Can we use `.Empty()` for these.

src/NzbDrone.Core/Applications/Whisparr/Whisparr.cs
```diff
@@ -84,15 +84,19 @@ public override List<AppIndexerMap> GetIndexerMappings()
...
 
-                var remoteIndexer = _whisparrV3Proxy.AddIndexer(radarrIndexer, Settings);
-                _appIndexerMapService.Insert(new AppIndexerMap { AppId = Definition.Id, IndexerId = indexer.Id, RemoteIndexerId = remoteIndexer.Id });
+                return;
             }
 
-            _logger.Trace("Skipping add for indexer {0} [{1}] due to no app Sync Categories supported by the indexer", indexer.Name, indexer.Id);
+            _logger.Trace("Adding indexer {0} [{1}]", indexer.Name, indexer.Id);
+
+            var radarrIndexer = BuildWhisparrIndexer(indexer, indexer.Protocol);
```
Qstick, line 96: care to fix this variable name while you are in here?

Qstick (changes requested):
Nice work here. Is the tag ignore going to cause some issues when we decide to fix #500 (I thinking of removing tags that were removed from the indexer during this sync)

PR #1498, opened 2023-03-01
Fixed: (UI) Try `event.composedPath()` for modern browsers

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/frontend/src/Search/SearchIndex.js
+++ b/frontend/src/Search/SearchIndex.js
@@ -84,6 +84,10 @@ class SearchIndex extends Component {
     }
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('keyup', this.onKeyUp);
+  }
+
   //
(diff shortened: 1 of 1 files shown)


frontend/src/Search/SearchIndex.js
```diff
@@ -192,7 +196,9 @@ class SearchIndex extends Component {
 
   onKeyUp = (event) => {
     const jumpBarItems = this.state.jumpBarItems.order;
-    if (event.path.length === 4) {
+    const path = event.composedPath ? event.composedPath() : event.path;
+
+    if (path.length === 4) {
```
Qstick, line 201: Can we just use below, it's what we've standardized on in other places, seems like most all modern browsers support

```
if (event.composedPath && event.composedPath().length === 4) {
```

Qstick (approved):

PR #1513, opened 2023-03-06
Remove the requirement for "search": [q]

### LINKED ISSUE #1512: Prowlarr should support disabling "q" search mode capability

Currently, if you write a custom indexer with a capability `search: []`, without `q`, you get the following error:

```
In search mode 'search' only 'q' parameter is supported and it's mandatory
```

However, I can't see any reason why this is necessary. And it seems that both [sonarr](https://github.com/Sonarr/Sonarr/blob/255f94786083b7935a2d7e4d1608d7698b3a3f15/src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs#L31) and [radarr](https://github.com/Radarr/Radarr/blob/b6b10d7c6f0d802775e98dbd7de0094b86b022e8/src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs#L25) actually check if `q` is supported before sending text query requests.

Also, currently even if you don't specify `q` under the movie or tvsearch capabilities, prowlarr seems to [populate them in the caps API](https://github.com/Prowlarr/Prowlarr/blob/e211436eb5375646e83daa3814d555b6120a0664/src/NzbDrone.Core/Indexers/IndexerCapabilities.cs#L282).

In my case, I have a tracker that only supports searching by TVDB or IMDB ID, no text queries. What's happening currently is, if an ID-based query has no results, sonarr / radarr fallback to using text queries, so I get a second, pointless query that will never give any results.

Specifying `search: []` (or even omitting the search field entirely) should be supported, so that "q" is no longer listed in "supportedParams" for the "caps" API. This way sonarr and radarr do not try to use text queries.

Also, the "caps" endpoint should not add `q` capabilities to movie / tvsearch and just reflect the yaml definition as written.

N/A

### Anything else?

N/A

scottjasso:
#### Database Migration
NO

#### Description
Currently, "search": [q] is a mandatory parameter. However this changes how sonarr and radarr search, forcing a mandatory text-search fallback if a query returns 0 results for e.g. tvdbId or imdbId. Some trackers don't support this, and radarr / sonarr seem to support ID-only based indexers.

This changes the code so that:
1) The ?t=caps endpoint only reflects the capabilities listed in the indexer definition (previously it implicitly added "q" for tvsearch, movie, etc, even if not specified in the yaml)
2) The "search: [q]" field is no longer required. Removing it implies your indexer doesn't support text-based search.

This is my first PR for Prowlarr, but I was able to build and run [most of] the tests locally just fine. I tested adding prowlarr as an indexer to sonarr, which worked well -- no text-search fallback after an empty tvdb result. However radarr had an issue: https://github.com/Radarr/Radarr/pull/8177

N/A

#### Todos
- [ ] Tests
- [ ] [Wiki Updates](https://wiki.servarr.com) -- **this is important, currently the wiki says the "search: [q]" bit is required.**
- [ ] Wait for https://github.com/Radarr/Radarr/pull/8177 to merge

* Fixes #1512

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Headphones/HeadphonesRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Headphones/HeadphonesRequestGenerator.cs
@@ -54,14 +54,14 @@ namespace NzbDrone.Core.Indexers.Headphones
             {
                 searchCriteria.SearchType = "search";
 
-                if (searchCriteria.SearchTerm.IsNotNullOrWhiteSpace() && capabilities.SearchAvailable)
+                if (searchCriteria.SearchTerm.IsNotNullOrWhiteSpace() && capabilities.SearchQueryAvailable)
                 {
                     parameters.Add("q", NewsnabifyTitle(searchCriteria.SearchTerm));
                 }
             }
             else
             {
-                if (searchCriteria.SearchTerm.IsNotNullOrWhiteSpace() && capabilities.MusicSearchAvailable)
+                if (searchCriteria.SearchTerm.IsNotNullOrWhiteSpace() && capabilities.MusicSearchQueryAvailable)
                 {
                     parameters.Add("q", NewsnabifyTitle(searchCriteria.SearchTerm));
                 }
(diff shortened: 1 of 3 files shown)


bakerboy448:
> From a Jackett point of view, we would not create an indexer that did not support a standard Torznab t=search,&q=term query. This is a Torznab mandated minimum requirement.
> We have to not only support those *arr apps that can do Torznab id search, but **all** torznab apps, which means being able to provide text search support as well.

_Originally posted by garfield69 in https://github.com/Prowlarr/Prowlarr/issues/1512#issuecomment-1456789031_


same applies to prowlarr for a Torznab spec viewpoint and UI searching;  convo to continue on #1512

bakerboy448:
https://github.com/Radarr/Radarr/commit/b3037248a231192adea8c384b713fdc83a04bf1d makes thhis unecessary

q is also required per torznab spec

PR #1515, opened 2023-03-07
Fixed: (UI) Reset header search offset value

scottjasso:
#### Database Migration
NO

#### Description
In the UI, if you search using the header input, the offset param continues to increase, even if you change the search terms. This gives unexpected search results (they're empty, since the backend doesn't like paged requests yet.) 

This fixes the search to reset the offset when you search via the header input.

#### Screenshot
1. Do a search
search1

2. Do another search, via header (not footer)
##### Before
search2
##### After
searchafter

#### Todos
- [ ] Tests

N/A


scottjasso:
> This fixes the issue with double requests for you too, one with `searchOffset=0` and the other with `searchOffset=100`?

Double requests? I don't think I was seeing that.

Qstick (approved):

PR #1522, opened 2023-03-11
Fixed: (PornoLab) Use correct download URL

selfhost-alt:
#### Database Migration
NO

#### Description
This indexer was previously using the `infoUrl` as the download URL, which cases downloads to fail when validating the torrent. The UI would show the download as started, but it would never show up in the torrent client, and Prowlarr logs would show a torrent parsing error (because of course an HTML page is not a valid torrent).

This change updates the `DownloadUrl` property to actually point at the torrent URL, enabling search results from this indexer to be sent to the torrent client for download.


Qstick:
Thanks, this is indeed a problem. There should have been a download override to the indexer to handle the extra request to get download url when the infoUrl was passed in (what Jackett does), but looks like that was left out. If your fix actually works it would save us that extra intermediate request and actually improve on that behavior. 

Can anyone test or confirm?

Qstick (approved):

PR #1525, opened 2023-03-12
fix(xthor): update details link

Tawagot0:
Example of link actually: https://xthor.tkdetails.php/?id=1

#### Database Migration
NO

#### Description
A few sentences describing the overall goals of the pull request's commits.

#### Todos
- [ ] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Xthor/XthorParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Xthor/XthorParser.cs
@@ -18,7 +18,7 @@ namespace NzbDrone.Core.Indexers.Definitions.Xthor
         {
             _settings = settings;
             _categories = categories;
-            _torrentDetailsUrl = _settings.BaseUrl.Replace("api.", "") + "/details.php?id={id}";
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Definitions/Xthor/XthorParser.cs
```diff
@@ -18,7 +18,7 @@ public XthorParser(XthorSettings settings, IndexerCapabilitiesCategories categor
         {
             _settings = settings;
             _categories = categories;
-            _torrentDetailsUrl = _settings.BaseUrl.Replace("api.", "") + "details.php?id={id}";
+            _torrentDetailsUrl = _settings.BaseUrl.Replace("api.", "") + "/details.php?id={id}";
```
Qstick, line 21:
```suggestion
            _torrentDetailsUrl = _settings.BaseUrl.Replace("api.", "").TrimEnd('/') + "/details.php?id={id}";
```

Add trimend to this to ensure it doesn't double in default case. This could probably be cleaned up more as this is only used iin one method, so doesn't really need to be a field defined on class

Qstick (approved):

PR #1562, opened 2023-04-01
Fixed: (Toloka) Add FreeleechOnly setting

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
@@ -323,6 +323,11 @@ namespace NzbDrone.Core.Indexers.Definitions
                 { "nm", term.IsNotNullOrWhiteSpace() ? term.Replace("-", " ") : "" }
             };
 
+            if (_settings.FreeleechOnly)
+            {
+                parameters.Add("sds", "1");
+            }
+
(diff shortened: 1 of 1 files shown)


Qstick (approved):

src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
```diff
@@ -408,6 +413,19 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                     MinimumSeedTime = 0
                 };
 
+                if (row.QuerySelector("img[src=\"images/gold.gif\"]") != null)
```
odayny, line 416: FYI: `images/authors.gif` also can be included here. Maybe I'll do id a bit later

odayny: Seems like a translation issue. In Ukrainian `завантажене` means `downloaded`, but it can mean `uploaded` as well in some contexts.

I don't have any link to FAQ (using this tracker since 2014 but never read the rules, shame on me) but if you open any release with this status there is exact same text as for gold (diff in the 1st word only):
```
Золотий реліз (завантажене не впливає на рейтинг)
Авторський реліз (завантажене не впливає на рейтинг)
```

PR #1579, opened 2023-04-05
Fixed: (AnimeBytes) Add `limit` and refactor parser

mynameisbogdan:
#### Database Migration
NO

#### Description
- Added limit depending on search query
- Added Freeleech Only setting
- Added Exclude RAW setting
- Added Exclude Hentai setting
- Set sorting to `grouptime desc`
- Refactor Synonymns to manage only nullable array of strings
- Change UploadTime type to string and parse with AssumeUniversal
- Refactor MST with additional 5 hours per GB 
- Use properties as list
- Fix when EnableSonarrCompatibility is true by not adding S01 to title for books, etc.
- Refactor episode/season padding
- Added new category parsing (PS2/PS3/SNES/NES/Switch/etc.)
- Order releases descending

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Text;
@@ -47,7 +48,7 @@ namespace NzbDrone.Core.Indexers.Definitions
 
         public override IParseIndexerResponse GetParser()
         {
-            return new AnimeBytesParser(Settings, Capabilities.Categories);
+            return new AnimeBytesParser(Settings);
         }
 
         protected override bool CheckIfLoginNeeded(HttpResponse httpResponse)
@@ -140,16 +141,20 @@ namespace NzbDrone.Core.Indexers.Definitions
             var searchUrl = $"{_settings.BaseUrl.TrimEnd('/')}/scrape.php";
 
             var term = searchCriteria.SanitizedSearchTerm.Trim();
+            var searchTerm = StripEpisodeNumber(term);
 
             var parameters = new NameValueCollection
             {
                 { "username", _settings.Username },
                 { "torrent_pass", _settings.Passkey },
+                { "sort", "grouptime" },
+                { "way", "desc" },
                 { "type", searchType },
-                { "searchstr", StripEpisodeNumber(term) }
+                { "searchstr", searchTerm },
+                { "limit", searchTerm.IsNotNullOrWhiteSpace() ? "50" : "20" }
             };
 
-            if (_settings.SearchByYear)
+            if (_settings.SearchByYear && searchType == "anime")
             {
                 var searchYear = ParseYearFromSearchTerm(term);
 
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
```diff
@@ -134,12 +135,17 @@ private IEnumerable<IndexerRequest> GetRequest(string searchType, string term, i
...
             var queryCollection = new NameValueCollection
             {
                 { "username", Settings.Username },
                 { "torrent_pass", Settings.Passkey },
+                { "sort", "grouptime" },
+                { "way", "desc" },
                 { "type", searchType },
-                { "searchstr", StripEpisodeNumber(term) }
+                { "searchstr", searchTerm },
+                { "limit", searchTerm.IsNotNullOrWhiteSpace() ? "50" : "15" }
```
proton-ab, line 148: I think you can even bump from 15 to 20 to match front-page coverflow.

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
```diff
@@ -235,300 +233,302 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
 
             var response = JsonConvert.DeserializeObject<AnimeBytesResponse>(indexerResponse.Content);
 
-            if (response.Matches > 0)
+            if (response.Matches == 0)
```
proton-ab, line 236: I do not think this is even necessary, `Groups` object is always present and in case of 0 matches it will simply be empty. Doing `foreach` on it should be enough to cover all cases without special checks.

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
```diff
@@ -56,6 +62,13 @@ protected override bool CheckIfLoginNeeded(HttpResponse httpResponse)
             return false;
         }
 
+        protected override Task<IndexerResponse> FetchIndexerResponse(IndexerRequest request)
+        {
+            var key = $"{request.HttpRequest.Url.FullUri}.{HashUtil.ComputeSha256Hash(Settings.ToJson())}";
+
+            return _searchCache.Get(key, () => base.FetchIndexerResponse(request), TimeSpan.FromMinutes(1));
```
proton-ab, line 69: I think bumping to 5 minutes would work better here. One have to also note that `/scrape.php` bases results off search index and that is updated every 5 minutes, so new torrents will take up to 5 minutes to show up in results.

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
proton-ab: > This sadly breaks the correct functionality in Prowlarr and Jackett, since it's assuming first is Japanese title, second Romaji title and third Alternative title.

This was always a wrong assumption to make though because we often put multiple titles delimited by `,` in `Alternate Title` field (legacy format doesn't really allow for unlimited number of aliases so that is somewhat of a nasty workaround, but works for search which is what we care about).

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
proton-ab: It's very unlikely that you'd see 20 new groups in 5 minutes window on AB and even then you are limiting RSS scraping to 15 minutes (or so I heard), so I don't see how that would affect anything.

Not really set on changing it though; just thought out I point out that search index we use is not updated instantly and always incurs delay of up to 5m.

src/NzbDrone.Core/Indexers/Definitions/AnimeBytes.cs
proton-ab: Mhm. Am I correct in assuming that these values are then used to send back to whoever asked Prowlarr/Jackett for data and not used when searching for data? If so I think the only solution here would be to revert the fix and deal with broken behavior as that is what API consumers been doing so far.

proton-ab: I've introduced new field `SynonymnsV2` and reveres old `Synonymns` to previous behavior.

```json
"SynonymnsV2": {
  "Japanese": "はじらいブレイク",
  "Romaji": "",
  "Alternative": "Bashful Break"
}
```

PR #1605, opened 2023-04-12
Fixed: (AnimeBytes) Cache result releases

mynameisbogdan:
#### Database Migration
NO


Qstick:
I don't have access to test this, but code looks good and approach is good like we discussed. Would you be up to hoping on discord to discuss quick? I want to test a few scenarios to make certain cache cleanup happens as it should after the expiration period and entries don't hang around in memory, we've had issues with that in past with Radarr.

Qstick (approved):

PR #1615, opened 2023-04-16
Fixed: UserAgent Parsing Ignores Dashes

buthed010203:
#### Database Migration
No

#### Description
The current UserAgent parser doesn't ignore dashes and requires a version number.
[Cross-seed's UserAgent](https://github.com/cross-seed/cross-seed/blob/1b65a0ec71ca823cfcfde04a52adcfc402df82d7/src/torznab.ts#L423) contains a dash and no version number and thus is not matched.

Also whats up with the `(?:\(.*\))?` non-capturing group at the end?

#### Todos
- Tests


buthed010203:
Fair enough but what about other tools that may not include version numbers?

buthed010203:
Fair enough

PR #1637, opened 2023-04-29
(Apprise) Change BaseUrl to ServerUrl

mynameisbogdan:
#### Database Migration
YES - 31

#### Description
1-1 with upstream.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/031_apprise_server_url.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/031_apprise_server_url.cs
@@ -1,0 +1,49 @@
+using System.Data;
+using Dapper;
+using FluentMigrator;
+using Newtonsoft.Json.Linq;
+using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(031)]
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Datastore/Migration/031_apprise_server_url.cs
```diff
@@ -0,0 +1,52 @@
...
+                        jsonObject.Add("serverUrl", jsonObject.Value<string>("baseUrl"));
+                        jsonObject.Remove("baseUrl");
+                    }
+
+                    settings = jsonObject.ToJson();
+                }
+
+                using var updateCmd = conn.CreateCommand();
+                updateCmd.Transaction = tran;
+                updateCmd.CommandText = "UPDATE Notifications SET Settings = ? WHERE Id = ?";
```
Qstick, line 45: Previously we've had issues with `?` parameters and Postgres, can we add a test here to make sure it's working on both?

Related conversation on the Sonarr PG PR <https://github.com/Sonarr/Sonarr/pull/5559/files#r1167634345>. I had don a condition for DB Type there, but Markus believe there to be a better way for both.

Qstick: Yep, looks good!

Qstick (approved):

PR #1641, opened 2023-04-30
New: (Rarbg) Add caching results

mynameisbogdan:
#### Database Migration
NO

#### Description
https://discord.com/channels/264387956343570434/264388019585286144/1102185483557154938

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Rarbg/Rarbg.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Rarbg/Rarbg.cs
@@ -5,14 +5,19 @@ using System.Net;
 using System.Threading.Tasks;
 using System.Web;
 using NLog;
+using NzbDrone.Common;
+using NzbDrone.Common.Cache;
 using NzbDrone.Common.EnvironmentInfo;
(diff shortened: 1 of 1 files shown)


Qstick:
Baker mentioned he can't comment because of the block 

Do we really want a setting here? It's messy to put this in user hands. Could we reduce to 10, do a cache like AB and also only use the second rate limit for search and not RSS? 

Just a few options (from Baker on Discord)

bakerboy448:
- The 429s I don't think are often triggered if they were then we could fix this prowlarr/jackett side
- the 520s we have no control over nor fix as it's their internal servers that are unable to handle the load.....unless we can control literally every single client hitting rarbg's servers which we can't :\

Suggest
- rss keep the 31s rate limit req but add a 30s ? cache like AB and similar to jackett; this should cut down on quite a few requests 
- keep normal searches at their existing limit

General comment: we're already doing exactly what rarbg asks; if their servers cannot handle their normal traffic load, that's on them

bakerboy448:
rebase ;)

bakerboy448 (approved):

Qstick (approved):
Let's give it a go

PR #1670, opened 2023-05-18
Fix validation for boolean query parameters

bakerboy448:
based on Sonarr 2fecd280016630d5afe2d60e52cbb52338c155b5

#### Database Migration
NO

#### Description
title


Qstick:
You may need `[FromBody]` and `[FromQuery]` on the two. And we'd need to test to make sure it's working in all scenarios then

- with and without the query parameter
- both versions of the query parameter

bogdan mentioned he'd dig in tomorrow.

Qstick:
Can we Squash

bakerboy448:
squished

PR #1703, opened 2023-05-30
Add some API attributes

mynameisbogdan:
#### Database Migration
NO


src/Prowlarr.Api.V1/openapi.json
Qstick, line 1: No need to delete this

Qstick (approved):

PR #1716, opened 2023-06-04
New: Use major version for installed version on Windows

### LINKED ISSUE #1715: Automatic updates do not reflect in Add/Remove Programs

With Prowlarr's (and the other *arr's) update system, the Add/Remove Programs (ARP) entry does not reflect the actually installed version. This can cause problems with things like winget or other system monitoring tools.

My preference is that the `DisplayName` is simply "Prowlarr" and the `DisplayVersion` is the correct, currently installed, version.

The `DisplayName` could just be "Prowlarr" and the `DisplayVersion` can be the major version only.

### Anything else?

Related to https://github.com/Sonarr/Sonarr/issues/5710

bakerboy448:
#### Database Migration
NO

#### Description
title
#### Todos
- Tests

* Fixes #1715

### DIFF
--- a/distribution/windows/setup/prowlarr.iss
+++ b/distribution/windows/setup/prowlarr.iss
@@ -9,6 +9,7 @@
 #define BaseVersion GetEnv('MAJORVERSION')
 #define BuildNumber GetEnv('MINORVERSION')
 #define BuildVersion GetEnv('PROWLARRVERSION')
+#define MajorBaseVersion Copy(BaseVersion, 1, Pos('.', BaseVersion)-1)
 
 [Setup]
 ; NOTE: The value of AppId uniquely identifies this application.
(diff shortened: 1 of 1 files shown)


bakerboy448:
> Sonarr/Sonarr@d606144

? Based on the changes done and existing downstream, the only material change needed is what's PR'd, no?

Qstick:
I think their intention with adding AppverName  was to override the default behavior of combination of name and ver. We should include that with ours.

distribution/windows/setup/prowlarr.iss
```diff
@@ -34,9 +34,10 @@ DisableReadyPage=True
 CompressionThreads=2
 Compression=lzma2/normal
 AppContact={#ForumsURL}
-VersionInfoVersion={#BaseVersion}.{#BuildNumber}
+VersionInfoVersion={#BaseVersion}
```
Qstick, line 37: BaseVersion here is actually the full version minus the build number. If you just want the actual major version (like Sonarr) this will need to be reworked a bit

https://github.com/Prowlarr/Prowlarr/blob/48cd1d9f6b55c71f42fab14871f14d7b15f4ed41/azure-pipelines.yml#L12

bakerboy448: fixup pushed; supposedly chatgpt thinks thatd work

Qstick: We'll see what the build artifact says

bakerboy448: chatgpt fixup! hopefully sorts it

PR #1724, opened 2023-06-08
Fixed: (Apps) Change the default sync level to Full Sync

mynameisbogdan:
#### Database Migration
NO

#### Description
Changed sync level to the recommended level.


Qstick (approved):
I'm fine with this now

PR #1739, opened 2023-06-22
Fix: (UI) Maintain search type and parameters on repeat search

### LINKED ISSUE #1317: When reusing search it does not include the paramters

### Current Behavior

When you try to run a search with the same exact query as another query in the history using the specific button for it, it doesn't actually include the parameters.

image

Here you can see I relaunched the search of Fate/Stay by pressing the little button, but both of my latest search didn't include the "Season 1" parameter.

### Expected Behavior

When using this button, the query should be exactly the same.

### Steps To Reproduce

- Create a search query with parameters (such as Season)
- Try to run the search manually by pressing the little button in the search history.
- It will not use the parameters

### Environment

```markdown
- OS: Windows 10 Pro
- Prowlarr: 1.0.1.2220 
- Docker Install: No
- Using Reverse Proxy: No
- Browser:Firefuck 102+
```

Master

### Trace Logs?

----------------------------------------------

mynameisbogdan:
#### Database Migration
YES - 034

#### Todos
- [x] Tests

* Fixes #1317

### DIFF
--- a/frontend/src/History/HistoryRow.js
+++ b/frontend/src/History/HistoryRow.js
@@ -8,9 +8,30 @@ import { icons } from 'Helpers/Props';
 import CapabilitiesLabel from 'Indexer/Index/Table/CapabilitiesLabel';
 import translate from 'Utilities/String/translate';
 import HistoryDetailsModal from './Details/HistoryDetailsModal';
+import * as historyDataTypes from './historyDataTypes';
 import HistoryEventTypeCell from './HistoryEventTypeCell';
 import HistoryRowParameter from './HistoryRowParameter';
 import styles from './HistoryRow.css';
+
+const historyParameters = [
+  { key: historyDataTypes.IMDB_ID, title: 'IMDb' },
+  { key: historyDataTypes.TMDB_ID, title: 'TMDb' },
+  { key: historyDataTypes.TVDB_ID, title: 'TVDb' },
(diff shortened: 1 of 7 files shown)


bakerboy448 (approved):
Seems fine here

frontend/src/History/HistoryRow.js
```diff
@@ -44,15 +44,56 @@ class HistoryRow extends Component {
...
+      'tvMazeId',
+      'season',
+      'episode',
+      'artist',
+      'album',
+      'label',
+      'track',
+      'year',
+      'genre',
+      'author',
```
Qstick, line 71: Can we do it without hard coding these? We already have to remember to update these in a few places on the backend that often get missed

bakerboy448: backend and front end - anyway to have these in a singular or twoish spots?

Qstick: Sure, I'll look at it when I have time.

Baker, lease avoid useless comments

Qstick: Haha, now we've gone to 3 extra places in UI to update them 😭

Qstick: Might be fine, I haven't looked close... Just saw more lists of parameters and head started spinning thinking of all the locations to add a new parameter

PR #1785, opened 2023-07-17
Set Transmission as SupportsCategories true

sentriz:
#### Database Migration

NO

#### Description

set transmission as SupportsCategories true. transmission does support categories (you can set a fallback category in prowlarr already in the client settings)

the reason for the change is that i'd like to be able to use the category mapping feature for transmission like i can for nzbget

#### Todos

- [ ] Tests

n/a


sentriz:
thanks makes sense, i'll have a look

src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
```diff
@@ -73,8 +73,21 @@ protected virtual OsPath GetOutputPath(OsPath outputPath, TransmissionTorrent to
             return outputPath + torrent.Name.Replace(":", "_");
         }
 
-        protected string GetDownloadDirectory()
+        protected string GetDownloadDirectory(string category)
         {
+            if (category.IsNotNullOrWhiteSpace())
```
sentriz, line 78: this feels a bit clunky but not sure of a better way to factor it without extra calls to `_proxy.GetConfig()`

sentriz:
Suit yourself, but it will take you a while.

> i'm a sucker for a false economy  

it builds now and testing locally, seems to be working well. tested the cases of

- no cateogry set
- some category set
- mapping of Audio to custom "audio" category  

in all cases transmission downloads to the expected place

wdyt? cheers @mynameisbogdan

sentriz:
PS the Waiting for OP label could be removed now i reckon

sentriz:
> I didn't had the time to test these changes, but at first glance I can tell you that there's no need to overload GetDownloadDirectory.

what would you suggest instead? I've not used c# before

sentriz:
that looks good but it looks to me like we will never use the custom download directory if the user has a download directory configured

image

we could hit the `if (Settings.Directory.IsNotNullOrWhiteSpace())` case right?

sentriz:
cool, suits me. i don't think i have that field explicity set anyway

sentriz:
@mynameisbogdan it's failing with some sort of screenshot issue. do you know what that's about? 
image

PR #1815, opened 2023-08-05
Fixed: (Indexers) Don't fetch releases when using unsupported capabilities

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
@@ -66,6 +66,20 @@ namespace NzbDrone.Core.Indexers
                 return Task.FromResult(new IndexerPageableQueryResult());
             }
 
+            var caps = GetCapabilities();
+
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -66,6 +66,18 @@ public override Task<IndexerPageableQueryResult> Fetch(MovieSearchCriteria searc
                 return Task.FromResult(new IndexerPageableQueryResult());
             }
 
+            var caps = GetCapabilities();
+
+            if ((!caps.MovieSearchImdbAvailable && searchCriteria.ImdbId.IsNotNullOrWhiteSpace()) ||
```
bakerboy448, line 71: If not movie imdb search is true AND if imdbid is populated

IMDb is supported true => false & id is populated; outcome is false, skip search

the negate seems to be wrong?

imdb is not supported inverts to true & IMDb has value; outcome is true and search?

(pre-coffee)

PR #1882, opened 2023-10-01
New: (Gazelle/OPS/RED) Prevent downloads without FL tokens

### LINKED ISSUE #1839: Option to prevent downloading without FreeLeech Token

The current behavior (at least in the referenced Orpheus config) is to automatically try downloading a torrent without a freeleech token if the token attempt fails.

This can be undesirable for many people, including myself, who prefer to only ever use FL tokens on high-difficulty ratio trackers. If Prowlarr proceeds to do this without user interaction, it can easily destroy ratios and the account standing with the tracker.

I would like to see a condition added to enforce Prowlarr to only ever attempt FL tokens on a torrent. This could look like a dropdown to replace the current "Use Freeleech Tokens" checkbox:

Use Freeleech Tokens:
- Never
- When available
- When size is above ___
- Always

N/A

### Anything else?

N/A

mynameisbogdan:
#### Database Migration
YES - 41

* Fixes #1839

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Gazelle/GazelleParser.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Gazelle/GazelleParser.cs
@@ -62,7 +62,7 @@ public class GazelleParser : IParseIndexerResponse
                     var isFreeLeech = torrent.IsFreeLeech || torrent.IsNeutralLeech || torrent.IsPersonalFreeLeech;
 
                     // skip releases that cannot be used with freeleech tokens when the option is enabled
-                    if (Settings.UseFreeleechToken && !torrent.CanUseToken && !isFreeLeech)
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Core/Indexers/Definitions/Gazelle/GazelleParser.cs
```diff
@@ -60,7 +60,7 @@ public virtual IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                 foreach (var torrent in result.Torrents)
                 {
                     // skip releases that cannot be used with freeleech tokens when the option is enabled
-                    if (Settings.UseFreeleechToken && !torrent.CanUseToken)
+                    if (Settings.UseFreeleechToken is (int)GazelleUseFreeleechTokens.Preferred or (int)GazelleUseFreeleechTokens.Required && !torrent.CanUseToken)
```
moly, line 63: I don't think you should skip these releases if use freeleech tokens is set to preferred. To me, "preferred" means I'm happy to see releases that tokens can't be used on too.

PR #1898, opened 2023-10-15
Fixed: Remove Defunct Tracker PirateTheNet

quorn23:
PTN closed down for good.

They had an official announcement to close down on the 14th October.

Aither offered to accept refugees. See /forums/topics/2535 on aither for more context.

#### Description
The definition is no longer needed as the tracker ceased all operation.


quorn23:
Redoing it properly

PR #1952, opened 2023-12-10
Fixed: Ignore terms with digits only in filter releases by query

mynameisbogdan:
#### Database Migration
NO

#### Description
Since some trackers seem to remove everything non-alphanumeric on a search, eg. `你想活出怎样的人生 2023` returns results for everything with `2023`, this change is trying to filter by at least one term with alphabetical characters.


bakerboy448:
Should/do we log at debug or trace that X releases were filtered out from Y indexer due to non-matching term?

Qstick:
Should this be handled per tracker? Since some trackers work fine with this

bakerboy448:
Is the scope wide enough to warrant a new RawSearch like flag?

PR #1963, opened 2023-12-22
New: Don't return API Keys and Passwords via the API

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/frontend/src/Store/Actions/indexerActions.js
+++ b/frontend/src/Store/Actions/indexerActions.js
@@ -225,7 +225,13 @@ export const reducers = createHandleActions({
     delete selectedSchema.name;
 
     selectedSchema.fields = selectedSchema.fields.map((field) => {
-      return { ...field };
+      const newField = { ...field };
+
+      if (newField.privacy === 'apiKey' || newField.privacy === 'password') {
+        newField.value = '';
+      }
+
+      return newField;
     });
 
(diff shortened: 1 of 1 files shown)


src/Prowlarr.Http/ClientSchema/SchemaBuilder.cs
```diff
@@ -28,13 +38,20 @@ public static List<Field> ToSchema(object model)
                 var field = mapping.Field.Clone();
                 field.Value = mapping.GetterFunc(model);
 
+                if (_configFileProvider.AuthenticationMethod is AuthenticationType.None or AuthenticationType.External &&
```
bakerboy448, line 41:
```suggestion
                if (_configFileProvider.AuthenticationMethod is AuthenticationType.None &&
```

punishing users of what Servarr Team pushes to be more secure than built in auth doesn't make sense (I.e. nginx auth, authelia, etc.)

Qstick: Yea, if I understand this, we are just allowing passwords open for those with auth explicitly enabled in app... vs all the other arr's where passwords are never shown (auth or not). This is actually opening things a bit from Sonarr/Radarr logic, not punishing users with External. I'm fine with this approach.

PR #2003, opened 2024-01-26
Change description of Tag field to match actual functionality

abcasada:
#### Database Migration
NO

#### Description
Fixes the description on the Tag field on the Edit Application page, because according to B on Discord in Prowlarr support channel:

"The flow is like this: no tags on application = sync all indexers, no matter what other tags the indexers have. app has at least one tag = only if common tag found."

#### Todos
- [ ] Tests

* None


bakerboy448 (approved):
Seems sane to me. But perhaps we should cover what occurs with no tags?

abcasada:
Ah, makes sense. I'm thinking of adding a second sentence. Does one of these sound right?

"If no tags are listed here, then all indexers (with matching categories) will be synced."
"If no tags are listed here, then no indexers will be prevented from syncing due to their tags."

bakerboy448:
If no tags are listed here, then no indexers will be prevented from syncing due to their tags. 

Seems succinct

abcasada:
> If no tags are listed here, then no indexers will be prevented from syncing due to their tags.
> 
> Seems succinct

This is done

bakerboy448 (approved):

PR #2065, opened 2024-03-12
Add download clients notice about sync to applications

mynameisbogdan:
#### Database Migration
NO


bakerboy448 (approved):
Defer to you on the UI specifics. Wording and alert seems fine to me

PR #2086, opened 2024-04-01
Update en.json

TheFiZi:
Sonarr, Radarr and Prowlarr all say "New update is available" but I can't determine which one it is without checking each one since I use the same notification method for each of them. This should resolve that.

#### Database Migration
NO

#### Description
I just want the update message to mention which service has the update available.

#### Todos
- [ ] Tests

* Fixes #XXXX


TheFiZi:
The Sonarr devs told me how the translations work. Will find another way to accomplish this.

PR #2100, opened 2024-04-10
New: Indexer Torrent Nostr

v0l:
#### Description
Implement nostr relay search for kind 2003 (Torrents) as seen on https://dtan.xyz

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/Nostr.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/Nostr.cs
@@ -1,0 +1,172 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


v0l:
> Hello, how is this supposed to work without a proxy download functionality via Prowlarr?

We're listing torrents on nostr so it should work the same as any of the other torrent index sites in prowler.

> For example? None of the arrs use documentation comments, so it's inherited.

`SupportsRss` `SupportsRedirect` `FollowRedirect` `Language` `LegacyUrls`

v0l:
> Did you tested your implementation? I'm asking because you're overriding the Download method with a NotImplemented which makes it invalid.

Yea this is a WIP, i just spent 1hr so far, going to finish it off tomorrow, looking for info on those props, thanks got it now.

I did some basic search tests so far with results, so everything seems fine so far

v0l:
Is it possible to first query the indexer to detect and to set `SupportsSearch`, not all relays support the search spec

PR #2174, opened 2024-07-09
New: Add `PreferMagnetUrl` to prefer magnet links over torrent file links

### LINKED ISSUE #1799: Add option to prefer magnet links for torrents

There are occasionally situations where a torrent name (so what shows up and is scored in Sonarr/Radarr for grab purposes) is different between the magnet link and the torrent file. This mainly affects anime as torrent/release naming vs file naming can vary significantly between various release groups, but could affect anything where these naming differences occur.

Example of a current VARYG release for Bleach from AnimeTosho:
Torrent name: Bleach Thousand-Year Blood War S01E16 The Fundamental Virulence 1080p NF WEB-DL AAC2.0 H 264-VARYG (Multi-Subs)
File name: Bleach.Thousand-Year.Blood.War.S01E16.The.Fundamental.Virulence.1080p.NF.WEB-DL.AAC2.0.H.264-VARYG.mkv
.torrent file name: Bleach.Thousand-Year.Blood.War.S01E16.The.Fundamental.Virulence.1080p.NF.WEB-DL.AAC2.0.H.264-VARYG.torrent
Magnet name: Bleach%20Thousand-Year%20Blood%20War%20S01E16%20The%20Fundamental%20Virulence%201080p%20NF%20WEB-DL%20AAC2.0%20H%20264-VARYG%20%28Multi-Subs%29 (url encoded, it's the same as the torrent name)

What happens in this scenario is that Sonarr will initially score the release high, but on import using the .torrent file it will be scored lower due to how the regex for VARYG is currently crafted. The regex is such that the release title must contain Multi-Subs to avoid their raw release as they have relatively poor naming. So it scores high initially (as it matches the regex via torrent name), but scores low on actual import. This has been noted to cause download loops as well. If added via magnet link (which Nyaa.si supports) the scoring is correct across the board and avoids issues.

Since the general recommendation is to not point automation at Nyaa.si it's not currently possible to automate magnet links from other sources that have magnet link support, such as AnimeTosho.

Similar to how Nyaa.si is implemented in Prowlarr currently, add an option to toggle prefer magnet links.

N/A

### Anything else?

N/A

Deathspike:
#### Database Migration
NO

#### Description
This pull request adds `PreferMagnetUrl` to indexer options. The aim of this option is to ensure that consuming applications will receive magnet links when available, instead of torrent file links. Applications that support Prowlarr (such as Radarr and Sonarr) should already support magnet links, as some providers do not provide torrent file links in the first place (e.g. TPB). This change thus leverages existing implementation for the handling of magnet links, tracking, and so on. Consuming applications should not be aware that Prowlarr may prioritize magnets over torrent file links.

This behavior only works when a magnet link is available for a release. If a magnet link is not available, the torrent file link is provided to applications as normal. Note that avoiding torrent file downloads may alleviate bandwidth and resource usage of the indexer, which could be favorable for said indexers. Additionally, because torrent files are often not downloaded with this option enabled, Prowlarr may avoid throttling or bans on certain indexer download endpoints (e.g. YTS).

The default value is off (retain current behavior).

Settings:

image

#### Todos
- [ ] Tests

* Fixes #1799


Deathspike:
I would love your feedback on the idea itself, but would also like to ask for guidance on the continuation for this PR. Specifically for tests and Wiki updates. What do you think is needed in those areas and how should I approach these? Thank you in advance!

Deathspike:
Understood. I'd like to avoid having to have a PR for both Radarr and Sonarr, so I'll investigate to see if there is a better way before I approach it that way. Thank you for the feedback. I will get back to this.

Deathspike:
I took another dive into the codebase and I believe that I found a better way to achieve the initial goal of preferring magnet links over torrent file links. When I was inspecting the situations that could occur and need to be supported, I learned that some indexers do not have torrent file links in the first place (e.g. TPB). That means that consuming applications already need to deal with proxied magnet links to function correctly. After tracing what happened in those situations, I stumbled on the implementation in [Radarr](https://github.com/Radarr/Radarr/blob/087f9e12aac522b958fdd3cd7671603ecd8153d1/src/NzbDrone.Core/Download/TorrentClientBase.cs#L146) and [Sonarr](https://github.com/Sonarr/Sonarr/blob/e97e5bfe8f3a7f01f7188f786eb7b51f1d8ef8b5/src/NzbDrone.Core/Download/TorrentClientBase.cs#L146) to handle proxied magnet links.

Well, it turns out that I was overthinking this feature. My earlier implementation of giving raw magnet links to consuming applications was unnecessary, and breaking tracking was easily avoided. The necessary behavior change is actually dead simple. Just swap the priority of `DownloadUrl` and `MagnetUrl` around in the `NewznabResults` XML builder. So, that's what I did. I have updated the PR details with a new title, description and screenshot to reflect these changes. Did I miss anything?

What do you think of the feature in this new state?

Deathspike:
Right. Let me know if you want me to change the PR from a global option to an indexer option.

Personally, I like choosing magnet links globally, but I understand the argument for an indexer option as well.

Deathspike:
May I ask what the status of this PR is? Should I take steps to keep moving this feature forward?

Deathspike:
For context, I spoke with bogdan about the status of this feature ~10 days on Discord.

> my main nitpick remains and that I don't think this should be a global option. now that some time has passed I'm 100% sure it should per indexer

I finally got around to it, so that's what I did in the latest set of changes. I moved the `PreferMagnetUrl` from a global option to a per-indexer option. This change requires a new column in `Indexers` table, so I added a migration and updated my initial post with the new 'Migration needed' status and updated description & details. 

The feature is certainly more versatile this way👍

frontend/src/Indexer/Edit/EditIndexerModalContent.js
```diff
@@ -116,6 +117,18 @@ function EditIndexerModalContent(props) {
                 />
               </FormGroup>
 
+              <FormGroup>
```
bakerboy448, line 120: Shouldn't this be disabled for non-torrent indexers?

Deathspike: You're absolutely right! I completely missed this. Fixed in [ac1181bbd93f57177eda05d83f38f9fde17ea286](https://github.com/Deathspike/Prowlarr/commit/ac1181bbd93f57177eda05d83f38f9fde17ea286). Thank you!

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -407,6 +407,8 @@
   "IndexerSettingsPackSeedTime": "Pack Seed Time",
   "IndexerSettingsPackSeedTimeIndexerHelpText": "The time a pack (season or discography) torrent should be seeded before stopping, empty is app's default",
   "IndexerSettingsPasskey": "Pass Key",
+  "IndexerSettingsPreferMagnetUrl": "Prefer Magnet URL",
+  "IndexerSettingsPreferMagnetUrlHelpText": "When enabled, this indexer will provide magnet links to applications if available with fallback to torrent links",
```
bakerboy448, line 411:
```suggestion
  "IndexerSettingsPreferMagnetUrlHelpText": "When enabled, this indexer will provide magnet links to applications if available with fallback to torrent links. If disabled, the torrent file will be used.",
```

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -407,6 +407,8 @@
   "IndexerSettingsPackSeedTime": "Pack Seed Time",
   "IndexerSettingsPackSeedTimeIndexerHelpText": "The time a pack (season or discography) torrent should be seeded before stopping, empty is app's default",
   "IndexerSettingsPasskey": "Pass Key",
+  "IndexerSettingsPreferMagnetUrl": "Prefer Magnet URL",
```
bakerboy448, line 410:
```suggestion
  "IndexerSettingsPreferMagnetUrl": "Prefer Magnet Link over Torrent URL",
```

bakerboy448 (commented):
Not set in stone but food for thought.

zakkarry (commented):
Engrish guy to the rescue.

I have a few options, I'll let you choose between them or mix and match parts or whatever however you prefer.

If you want a second run at what you like from this and that, just let me know.

src/NzbDrone.Core/Localization/Core/en.json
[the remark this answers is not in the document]
bakerboy448: Suggest this

src/NzbDrone.Core/Localization/Core/en.json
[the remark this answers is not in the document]
bakerboy448: Suggest this for helptext

src/NzbDrone.Core/Localization/Core/en.json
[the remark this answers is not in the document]
zakkarry: Would specifying (similar to how I did in another with "for grabs") be able to be done concisely with those specifics?

It seems like a very specific use case and I'm not sure what common language the user would be familiar with to convey that without making it significantly longer.

PR #2221, opened 2024-09-09
return error status if unsupported capabilities are used.

jaype87:
#### Database Migration
NO

#### Description
Report errors instead of empty results in case of unsupported capabilities used in search.

I wanted to check with you if something like this is generally acceptable and made a rudimentary implmentation.

#### Todos
- [ ] Tests
- [ ] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json) - No UI impact
- [ ] [Wiki Updates](https://wiki.servarr.com) - I don't think there's anything to update.

### DIFF
--- a/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
+++ b/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
@@ -7,6 +7,7 @@ using NzbDrone.Common.Instrumentation.Extensions;
 using NzbDrone.Core.Exceptions;
 using NzbDrone.Core.Indexers;
 using NzbDrone.Core.Indexers.Events;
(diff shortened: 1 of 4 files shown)


jaype87:
Yeah, this is just a quick hack to demonstrate the concept. Happy to change the implementation, but since I'm not very familiar with the code base, I'd need some pointers oon how to do this without affecting other use cases.

For some background: LazyLibrarian doesn't trust reported capabilities. So it'll try the book search with author and title. If it get's an error, it falls back to general search. But since Prowlarr just returns an empty list, this fallback doesn't happen.

jaype87:
I'm not sure of the historic reasons, I only recently started contributing there. I'll bring it up there. Thanks for your feedback.

PR #2235, opened 2024-09-28
fix: fix broken BeyondHD API

tvcsantos:
#### Database Migration
NO

#### Description
Currently BeyondHD API is broken and is not filtering properly "freelech" or "limited" torrents. This PR addresses that issue, making sure that in case of wrong results they are filtered anyway after getting the response.

#### Todos
- [x] Tests


src/NzbDrone.Core/Indexers/Definitions/BeyondHD.cs
```diff
@@ -274,6 +276,18 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                     flags.Add(IndexerFlag.Internal);
                 }
 
+                // BHD API is broken on limited filter
```
bakerboy448, line 279:
```suggestion
                // ToDo: Remove when BHD Fixes API. BHD API is broken on limited filter
```

any ETA from staff when they may have it fixed up?

tvcsantos: No ETA on their side for now.

PR #2278, opened 2024-11-14
Ignore "Percona Distribution" in Postgres version string

aronsky:
Drop the ` - Percona Distribution` suffix before passing the version string to `System.Version`.

#### Description
I tried using Sonarr with Percona Everest (a k8s operator that provisions various databases, including PostgreSQL). However, Percona Everest provisions a custom version of PostgreSQL, that reports its version as "16.3 - Percona Distribution". This causes an exception when that version is passed to the `System.Version` constructor.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Database.cs
+++ b/src/NzbDrone.Core/Datastore/Database.cs
@@ -52,7 +52,7 @@ namespace NzbDrone.Core.Datastore
             {
                 using var db = _datamapperFactory();
                 var dbConnection = db as DbConnection;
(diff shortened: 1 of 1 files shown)


bakerboy448 (changes requested):
- build falls
- this is a very hacky workaround
- needs tests
- likely should go upstream to sonarr first

aronsky:
Build failure doesn't seem related to the code change.

bakerboy448 (changes requested):
Concur with suggested concept change from Markus to parse out the version

https://github.com/Sonarr/Sonarr/pull/7385#issuecomment-2476490516

PR #2304, opened 2024-12-11
fix: norwegian characters in NorBits searchTerm

eldarrarr:
#### Description
Prowlarr handles Norwegian special characters very well, by searching both with the actual character and with the non special variation (e.g. `ø` becomes `oe`).

The special character is encoded, as it should be. 
The issue however, is that the search query in the URL for NorBits requires _the actual special character_.

So if someone releases something on NorBits and uses ø in the name instead of `oe`, it's not searchable by Prowlarr, because it searches with `%F8` which is interpreted literally on NorBits.

This fixes this issue by replacing the encoded version of the Norwegian characters back to their non-encoded version, after the rest of the searchTerm has been encoded (which is needed, because `+` is still needed for space etc.).


s0up4200:
@zodihax Can you provide an example with one that currently fails? The ID should be enough.

s0up4200:
Just trying to replicate the issue. Take 224780 as an example. I can search for the release name and it finds it just fine. Even going into the History tab afterwards to click the link for the query works.

PR #2305, opened 2024-12-12
fix: norwegian characters in NorBits searchTerm

zodihaxx:
#### Description
Prowlarr handles Norwegian special characters very well, by searching both with the actual character and with the non special variation (e.g. `ø` becomes `oe`).

The special character is encoded, as it should be.
The issue however, is that the search query in the URL for NorBits requires the actual special character.

So if someone uploads something on NorBits and uses `ø` in the name instead of `oe`, it's not searchable by Prowlarr, because it searches with `%F8` which is interpreted literally on NorBits.

This PR fixes this issue by replacing the encoded version of the Norwegian characters back to their non-encoded version, after the rest of the searchTerm has been encoded (which is needed, because + is still needed for space etc.).

An example ID is `223078`.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/NorBits.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/NorBits.cs
@@ -191,6 +191,14 @@ public class NorBitsRequestGenerator : IIndexerRequestGenerator
         else if (!string.IsNullOrWhiteSpace(term))
         {
             searchTerm = "search=" + term.UrlEncode(Encoding.GetEncoding(28591));
+            
+            searchTerm = searchTerm
+            .Replace("%E6", "æ")
+            .Replace("%F8", "ø")
+            .Replace("%E5", "å")
+            .Replace("%C6", "Æ")
+            .Replace("%D8", "Ø")
+            .Replace("%C5", "Å");
         }
 
         searchUrl += "?" + searchTerm + "&" + parameters.GetQueryString();


zodihaxx:
Please re-run the checks? The exact same code passed the checks in the last PR 🤔

zodihaxx:
I think maybe I'm understanding this a bit more, with more testing.

It seems like NorBits expects UTF8, which is how I interpret my testing, and thus edited my PR.
So now it encodes to UTF8 instead of ISO, which should hopefully work.

zodihax:
@s0up4200 did last time, maybe you can help out this time too?

s0up4200:
> @s0up4200 did last time, maybe you can help out this time too? 

Can't today, but will try to remember testing it tomorrow evening!

s0up4200:
@zodihax I find that release (223078) by searching for the release title in Prowlarr just fine. `Vårt.L....`

zodihax:
You must be on another version of Prowlarr or something? Did you search with the entire name including both the Norwegian characters?
I'm on version `1.27.0.4852`.

The reason I mentioned that ID as an example, is because someone other than me has the same issue and posted about it in the forums -- with that ID specifically. You'll see the post in recent activity on NorBits.

s0up4200:
I used the full release name yeah. With both `Å` and `Ø` present.

v1.28.0.4862
https://i.imgur.com/BIetgK3.png

zodihax:
Docker version? Like what could be different from my (ane the guy in the forum’s setup) than yours? In the link that you see in history in Prowlarr, do you see the characters encoded as ISO or UTF8?

s0up4200:
`Docker version 27.3.1, build ce12230`
Prowlarr package version `1.28.0.4862-ls203` by [linuxserver.io](https://www.linuxserver.io/)

https://i.imgur.com/nINhrZs.png

zodihax:
You're on `nightly`. 😅

But I upgraded to nightly image too and tried, still no results.

Then I changed the indexer and selected "Use Full Search" at which point I got the results.

So if you uncheck "Use Full Search" you should be able to replicate the bug.

s0up4200:
Can confirm it doesn't find it with `Use Full Search` set to disabled.

zodihax:
@mynameisbogdan finally the bug is reproduced by someone else 😊

s0up4200:
> > Can confirm it doesn't find it with `Use Full Search` set to disabled.
> 
> 
> 
> But the current PR fixes it for you? If you can't build and run from source, I can do a docker image with a PR build later.

If you can build an image that would be very much appreciated 👍

s0up4200:
Can confirm the search returns the release with `Use Full Search` set to disabled now. And with it set to enabled for that matter.

PR #2350, opened 2025-02-17
New: Support for exclusive indexer flag

mynameisbogdan:
#### Database Migration
NO

#### Description
Support for exclusive indexer flag where the release should not be uploaded or cross-seeded in other places.

### DIFF
--- a/src/NzbDrone.Core/Indexers/IndexerFlag.cs
+++ b/src/NzbDrone.Core/Indexers/IndexerFlag.cs
@@ -63,6 +63,7 @@ namespace NzbDrone.Core.Indexers
         }
 
         public static IndexerFlag Internal => new ("internal", "Uploader is an internal release group");
+        public static IndexerFlag Exclusive => new ("exclusive", "An exclusive release that must not be uploaded anywhere else");
         public static IndexerFlag FreeLeech => new ("freeleech", "Download doesn't count toward ratio");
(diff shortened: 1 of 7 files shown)


src/NzbDrone.Core/Indexers/IndexerFlag.cs
```diff
@@ -63,6 +63,7 @@ public override bool Equals(object obj)
         }
 
         public static IndexerFlag Internal => new ("internal", "Uploader is an internal release group");
+        public static IndexerFlag Exclusive => new ("exclusive", "An exclusive release should not be uploaded anywhere else");
```
bakerboy448, line 66:
```suggestion
        public static IndexerFlag Exclusive => new ("exclusive", "An exclusive release shall not be uploaded anywhere else");
```

bakerboy448 (approved):

src/NzbDrone.Core/Indexers/IndexerFlag.cs
zakkarry:
```suggestion
        public static IndexerFlag Exclusive => new ("exclusive", "An exclusive release that must not be uploaded anywhere else");
```

zakkarry: @mynameisbogdan co-author!

zakkarry: Now requires another approval ahahaha

Coauthor would make it approvable :P

PR #2395, opened 2025-05-12
New: Updated all newznab indexers to be redirect

RetroDadson:
#### Database Migration
YES - 043

#### Description
Following on from https://github.com/Prowlarr/Prowlarr/commit/411e96ef2a9d585f2bca651f7e1ebab270432aae there are still indexers requesting that we have this set on existing instances due to account sharing. This updates all existing Newznab indexers to be set to redirect.

#### Todos
- [x] Tests

### DIFF
--- a/src/NzbDrone.Core.Test/Datastore/Migration/043_newznab_indexers_enable_redirectFixture.cs
+++ b/src/NzbDrone.Core.Test/Datastore/Migration/043_newznab_indexers_enable_redirectFixture.cs
@@ -1,0 +1,192 @@
+using System;
+using System.Linq;
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Datastore.Migration;
+using NzbDrone.Core.Test.Framework;
+
+namespace NzbDrone.Core.Test.Datastore.Migration
+{
+    [TestFixture]
+    public class newznab_indexers_enable_redirectFixture : MigrationTest<newznab_indexers_enable_redirect>
+    {
+        [Test]
+        public void should_update_redirect_setting_to_true_if_false()
+        {
+            var db = WithMigrationTestDb(c =>
+            {
+                c.Insert.IntoTable("Indexers").Row(new
+                {
+                    Id = 1,
+                    Name = "Test",
+                    Implementation = "Newznab",
+                    Settings = "{\"baseUrl\":\"https://example.com\",\"apiKey\":\"testapikey\"}",
+                    ConfigContract = "NewznabSettings",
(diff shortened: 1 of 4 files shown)


RobinDadswell:
@mynameisbogdan are you able to take a look at this and see if you can think of a better way to force redirect to be true for usenet indexers? If not I will get this merged later today :)

RobinDadswell:
@mynameisbogdan the reason that we are looking to force redirect to on is because we are being asked to by various Indexers, is there any reason to not force it to on?

RobinDadswell:
> I'm aware there's a new indexer requesting redirect to be enabled, but due to the nature of the newznab in Prowlarr we can't handle this property per indexer unless we implement validation for redirect per hostname. Which will break it when an indexer will demand to not be named in code.
> 
> Can't think of anything right now, but we might find out in the future if a new indexer with a crappy implementation comes along with some measure which will prevent downloading and that's why I think it we should leave the option user-configurable for now.
> 
> Setting SupportsRedirect to false might do the opposite anyway, just because I assume you want to disable the checkbox in UI.
> 
> https://github.com/Prowlarr/Prowlarr/blob/903b86b9a215b54f72e7261739deb36565d11cc3/src/Prowlarr.Api.V1/Indexers/NewznabController.cs#L266

Ahh I didn't find this bit, 

Yes I wanted to drop the user configurable bit - I am very aware that it's not just one indexer but all indexers. I will have a look at the Newznab controller to just use the "redirect" setting and ignore the SupportsRedirect piece. 

Ultimately usenet indexers are starting to require it enmasse, we can't do hostnames based on the ones that demand not to be named and so we need to work around it another way

RobinDadswell:
I did start looking at purely doing it via validation, the issue is that redirect isn't a setting that we could validate on in it's current state as it's passed through via the UI not via the standard settings page... I will look to see if there is a better way around it

RobinDadswell:
> Ah, you're right. And the arrs validate the settings only anyway in
> 
> https://github.com/Prowlarr/Prowlarr/blob/8cb59c35fbc76db1b3b9f713041e77c512b03c8a/src/Prowlarr.Api.V1/ProviderControllerBase.cs#L276
> 
> Let me look into the next days as well, as I guess enabling it in backend would allow us to remove the db migration.
> 
> Implementing validation directly on providers is another solution to this, I dare say. 👍

I think we still need the DB migration to pick everything up that's already setup, but yeah I looked at the providers as an alternative for this but wasn't overly sold on it when I first looked

CicerBro:
Any updates on this? 👀

PR #2454, opened 2025-07-30
Add optional option for sessionId with flaresolverr

doodz:
#### Database Migration
 NO

#### Description
Add optional option for sessionId with flaresolverr
The session ID that you want to be assigned to the instance.
If isn't set a random UUID will be assigned from Flaresolver.

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/IndexerProxies/FlareSolverr/FlareSolverr.cs
+++ b/src/NzbDrone.Core/IndexerProxies/FlareSolverr/FlareSolverr.cs
@@ -101,7 +101,7 @@ namespace NzbDrone.Core.IndexerProxies.FlareSolverr
 
             var url = request.Url.ToString();
             var maxTimeout = Settings.RequestTimeout * 1000;
-
+            var sessionId = Settings.SessionId;
             var proxySettings = _proxySettingsProvider.GetProxySettings();
             var proxyUrl = proxySettings != null ? GetProxyUri(proxySettings) : null;
 
@@ -119,6 +119,7 @@ namespace NzbDrone.Core.IndexerProxies.FlareSolverr
                     Cmd = "request.get",
                     Url = url,
                     MaxTimeout = maxTimeout,
+                    SessionId = sessionId,
                     Proxy = requestProxy
                 };
             }
@@ -142,6 +143,7 @@ namespace NzbDrone.Core.IndexerProxies.FlareSolverr
                             ContentLength = null
                         },
                         MaxTimeout = maxTimeout,
+                        SessionId = sessionId,
                         Proxy = requestProxy
                     };
                 }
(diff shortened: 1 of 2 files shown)


bakerboy448:
- why is this just now required after years and years of flaresolverr support?
- what benefit does this add for a dying application (flaresolverr) and a losing battle (CF is winning) for Trackers that make the conscious and deliberate choice to be anti-automation and not build a UI / place CF in front of their site?

ngotchac:
Too bad this got closed, I was looking for this feature. I don't understand the pushback, it's +8 LOC and speeds-up flaresolverr significantly. I 100% understand the frustration of having to use and/or support flaresolverr, but if the intergration is bound to not receive any updates ever, shouldn't it simply be removed?

ilike2burnthing:
This is either going to open a FlareSolverr session that is never closed (obviously a terrible idea), or open a session for a single use and then close it immediately after, like currently, but now with a specific ID (making it completely pointless).

ngotchac:
Unless I'm misunderstanding something, I would use it as such: start a session on Flaresolverr when starting Prowlarr, and make Prowlarr use this specific session. The session would be stopped with Prowlarr as well. I wouldn't have session leaks, and all requests will be much faster, as I wouldn't need to redo the Cloudflare challenge most of the time

ilike2burnthing:
> The session would be stopped with Prowlarr as well.

There's nothing in this PR that would do that.

> I wouldn't have session leaks

?

> and all requests will be much faster, as I wouldn't need to redo the Cloudflare challenge most of the time

Once the challenge is solved, the cookies are saved by Prowlarr and reused for subsequent requests until the cookie expires. specifying a session ID wouldn't change this.

ngotchac:
> There's nothing in this PR that would do that.

I didn't  say it was part of this PR. But w/ this PR, I'd put in place some bash scripts to do that, it's pretty easy to do.

Not sure what's unclear about "I wouldn't have session leaks" -> as explicited above, I can manage that myself.

> Once the challenge is solved, the cookies are saved by Prowlarr and reused for subsequent requests until the cookie expires. specifying a session ID wouldn't change this.

Indeed, it improves the request duration. However, it still takes time for Flaresolverr to setup everything, even with the correct cookie. I just tested, and w/o a session, but w/ the right cookie, it takes ~1.5s. However, with a session, after the challenge, it takes ~250ms. It's not that much for a single request, but can add up.

Again, I might not know all the implications of merging this PR, but it seems to me that for 8 lines of changes, it's a nice addition for not so much added maintenance?

ilike2burnthing:
Anything that requires users to write and run their own scripts isn't going to be added.

ngotchac:
Any reason for that? It can be hidden as an "advanced settings". W/ the same reasoning, why expose the API then? Surely users will have to write "scripts" to use it.

ilike2burnthing:
There are users who just about manage to run programs on Windows, there are others who run it on glorified potatoes; in neither case is it a good idea to give them a setting that just leaves Chrome/Chromium constantly running unless they write and run scripts, all for the sake of the odd request being a second faster.

ngotchac:
All right, I guess there's no way to convince you otherwise. Thanks for taking the time to answer anyway! :)

PR #2487, opened 2025-09-01
fixed:(unit3d) handle missing torrent hash in infodict

Audionut:
#### Database Migration
NO

#### Description
torrent files from unit3d based sites, v9.1.6 and higher, no longer contain hash in the info dict. This pr returns an empty string on hash parsing failure, relying on hash from the release info instead


bakerboy448:
Thanks, but

- CI failed
- Prowlarr changes are not required for what is a UNIT3D Cardigann defintion update. This is not the proper fix.
- as previously mentioned, Prowlarr/Indexers: definitions just need to be updated from jackett which no one from the community seems willing/able to run a sync script or PR the CI branch

PR #2500, opened 2025-09-13
Fixed: (BroadcasTheNet) Improve daily episode searching

thisaccountwillmakeprs:
#### Database Migration
NO

#### Description
Named episodes such as `1970.01.01 - Episode Title` were not previously picked up. This should fix that.

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
@@ -140,7 +140,7 @@ namespace NzbDrone.Core.Test.IndexerTests.BroadcastheNetTests
             query.Tvrage.Should().BeNull();
             query.Search.Should().BeNull();
             query.Category.Should().Be("Episode");
-            query.Name.Should().Be("2023.01.03");
+            query.Name.Should().Be("2023.01.03%");
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


bakerboy448:
What changed in their api to need this? There was a lot of rework and testing done a few years ago.  Also it is an episode parameter not an episode query. Is BTN no longer properly handling episode parameters the same?

https://github.com/Prowlarr/Prowlarr/commit/5fdc6ee25dc37fc09674f7d925a869b992a9e5a1?w=1

https://github.com/Prowlarr/Prowlarr/pull/1601

thisaccountwillmakeprs:
> What changed in their api to need this? There was a lot of rework and testing done a few years ago. Also it is an episode parameter not an episode query. Is BTN no longer properly handling episode parameters the same?
> 
> [5fdc6ee?w=1](https://github.com/Prowlarr/Prowlarr/commit/5fdc6ee25dc37fc09674f7d925a869b992a9e5a1?w=1)
> 
> #1601

I actually took a look at that linked PR prior to writing mine and that seems to be where this problem was introduced. We already append a % to non daily episodes for what I'm assuming is the same reason. I don't know that they actually changed anything, not every show has group names with episode titles as far as I can tell. Take 20/20 as an example, the latest episode has a named group so the old search wouldn't pick it up: `"GroupName": "2025.09.05 - Idaho Justice"`.

Here's a tiny python script that demonstrates this change working:
```py
import requests
from json import dumps

key = "apikey"

def request(method, params):
    req = requests.post("https://api.broadcasthe.net", json={"method": method, "jsonrpc": "2.0", "id": 1, "params": params})
    json = req.json()
    print(dumps(json, indent=4))
    return int(json["result"]["results"])

assert request("getTorrents", [key, {"Category":"Episode","Name":"2025.09.05","Tvdb":"72289"}]) == 0 # Old
assert request("getTorrents", [key, {"Category":"Episode","Name":"2025.09.05%","Tvdb":"72289"}]) > 0 # New
```

> Also it is an episode parameter not an episode query

Right but that gets converted into a query down the road

thisaccountwillmakeprs:
It actually appears that B missed this when fixing it for other episode types last year https://github.com/Prowlarr/Prowlarr/commit/a85406e3b7871fac37b9bb506d2d83ba15848d4e

bakerboy448 (approved):

Issue #2510, opened 2025-09-30
RevolutionTT URL has changed and needs to be updated


renegadepixels:
### I attest that there is not an existing issue for this?

- [x] I have searched the existing open and closed issues

### I attest this is not related to a Cardigann YML Indexer.

- [x] I confirm this is not related to a Cardigann YML Indexer

### Current Behavior

Revolutiontt.me is now hosted at RevoTT.me so the URLs need to be updated.

### Expected Behavior

Expect indexer to connect

### Steps To Reproduce

Try to add RevolutionTT as an indexer and it will not connect.

### Environment

```markdown
- OS: Windows
- Prowlarr: 2.0.5.5160
- Docker Install: No
- Using Reverse Proxy: No 
- Browser: Chrome/Firefox
```

### What branch are you running?

Master

### Trace Logs?

N/A I believe, if I am wrong and you need logs let me know...

### I attest that Trace Logs have been provided as applicable. Reports will be closed if the required logs are not provided.

- [x] I attest that I have read and followed the steps in the wiki link above and provided the required trace logs - the logs contain `trace` - that are relevant and show this issue.

bakerboy448:
https://github.com/Prowlarr/Prowlarr/commit/2f7d821d45926ec7811eb61296772fa224d0968f

PR #2511, opened 2025-09-30
Fixed: (RevolutionTT) New Domain

zbowling:
#### Database Migration
NO

#### Description
RevolutionTT switched domain to revott.me. 

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/RevolutionTT.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/RevolutionTT.cs
@@ -24,6 +24,7 @@ namespace NzbDrone.Core.Indexers.Definitions
         public override string Name => "RevolutionTT";
 
         public override string[] IndexerUrls => new[] { "https://revott.me/" };
+        public override string[] LegacyUrls => new[] { "https://revolutiontt.me/" };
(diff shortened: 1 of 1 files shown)


bakerboy448 (changes requested):
We can't just add the new url.
We also need to move the now defunct url to the legacy url for auto-migration

src/NzbDrone.Core/Indexers/Definitions/RevolutionTT.cs
```diff
@@ -23,7 +23,7 @@ public class RevolutionTT : TorrentIndexerBase<UserPassTorrentBaseSettings>
     {
         public override string Name => "RevolutionTT";
 
-        public override string[] IndexerUrls => new[] { "https://revolutiontt.me/" };
+        public override string[] IndexerUrls => new[] { "https://revott.me/" };
```
ilike2burnthing, line 26:
```suggestion
        public override string[] IndexerUrls => new[] { "https://revott.me/" };
        public override string[] LegacyUrls => new[] { "https://revolutiontt.me/" };
```

ilike2burnthing:
Resolves #2510

zbowling:
Aww sweat! Thanks!

PR #2575, opened 2026-01-02
Fixed: Form data encoding for non-UTF8 indexers

### LINKED ISSUE #1871: ISO 8859-1 special and accented characters not supported

### Current Behavior

For an indexer using ISO 8859-1 encoding, logins fail when using a password which contains special or accented characters, see rows Ax to Fx here - https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Code_page_layout

### Expected Behavior

Works, as in Jackett.

### Steps To Reproduce

1. have password containing special or accented character for ISO 8859-1 encoded tracker (e.g. Deildu, Genesis-Movement, TribalMixes)
2. try to add indexer
3. fail with 'wrong credentials' error, despite them being correct

### Environment

```markdown
- OS: Windows 10 Sandbox
- Prowlarr: 1.8.6.3946
- Docker Install: No
- Using Reverse Proxy: No
- Browser: MS Edge
```

Master

### Trace Logs?

N/A - just normal failed login logs

### Trace Logs have been provided as applicable. Reports may be closed if the required logs are not provided.

- [X] I have read and followed the steps in the wiki link above and provided the required trace logs - the logs contain `trace` - that are relevant and show this issue.

segalll:
#### Database Migration
NO

#### Description
When logging into indexers that use non-UTF8 encodings (like RuTracker with windows-1251), usernames or passwords containing non-ASCII characters would fail authentication. The issue was that Uri.EscapeDataString always percent encodes using UTF-8, ignoring the encoding set on the request builder.

This changes HttpRequestBuilder.ApplyFormData to use the encoding-aware UrlEncode method instead, so form data is properly encoded according to what the indexer expects.

Also added explicit SetEncoding() calls to the login flows of affected indexers (RuTracker, PornoLab, TorrentBytes, PreToMe, FunFile) to ensure the correct encoding is used when building form parameters.

#### Todos
- [x] Tests

* Fixes #1871

### DIFF
--- a/src/NzbDrone.Common.Test/Http/HttpRequestBuilderFixture.cs
+++ b/src/NzbDrone.Common.Test/Http/HttpRequestBuilderFixture.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Text;
 using FluentAssertions;
 using NUnit.Framework;
 using NzbDrone.Common.Http;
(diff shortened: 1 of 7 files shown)


mynameisbogdan:
Looks reasonable. Would be possible for a few tests to ensure this works as intended and won't be broken by mistake in the future?

PR #2600, opened 2026-02-10
Fixed: Use FlareSolverr response body instead of re-requesting

### LINKED ISSUE #2561: Prowlarr not using Proxy Response Content for YTS

### Current Behavior

Byparr is returning the YTS JSON just fine, trace log shows a success response and the correct/solved Response Content JSON being provided to prowlarr.
Prowlarr seems to then completely ignore the correct/solved Response Content and sends another request without Byparr which then obviously returns a Cloudflare issue which Prowlarr then registers as a cloudflare issue for the whole index, despite it already having received the correct/solved Response Content from Byparr.

### Expected Behavior

Byparr returns correct/solved Response Content JSON.
Prowlarr then uses this JSON and doesn't submit another HttpClient|Req: [GET] request.
Prowlarr doesn't register the index request as failed since it has already received the correct/solvved Response  Content JSON.

### Steps To Reproduce

YTS Index with Byparr proxy tag

### Environment

```markdown
- OS: Windows 10
- Prowlarr:2.3.0.5236
- Docker Install: 4.48.0 (207573)
- Using Reverse Proxy: No
- Browser: Chrome Version 142.0.7444.176
```

Master

### Trace Logs?

[Prowlarr Trace Log - YTS Issue.txt](https://github.com/user-attachments/files/23892730/Prowlarr.Trace.Log.-.YTS.Issue.txt)

quaintops:
Fixes #2561

### Summary

- Use FlareSolverr's solved HTML response body directly instead of discarding it and making a second HTTP request with cookies
- The second request fails because Cloudflare's `cf_clearance` cookie is bound to the solver's TLS fingerprint, which .NET's HttpClient cannot replicate
- Fall back to cookie-based retry only when FlareSolverr returns no body (preserving original behavior)
- Fix error status code reference (`response.StatusCode` → `flaresolverrResponse.StatusCode`)

### Test plan

- [ ] Configure FlareSolverr/Byparr as an indexer proxy in Prowlarr
- [ ] Add a Cloudflare-protected indexer (e.g. LinuxTracker) tagged with the proxy
- [ ] Verify the indexer loads successfully using the FlareSolverr response body
- [ ] Verify indexers **not** behind Cloudflare still work normally (no regression)
- [ ] Verify the fallback path works when FlareSolverr returns an empty body

https://github.com/Prowlarr/Prowlarr/issues/2561#issuecomment-3879907891


src/NzbDrone.Core/IndexerProxies/FlareSolverr/FlareSolverr.cs
```diff
@@ -58,7 +58,7 @@ public override HttpResponse PostResponse(HttpResponse response)
 
             if (flaresolverrResponse.StatusCode != HttpStatusCode.OK && flaresolverrResponse.StatusCode != HttpStatusCode.InternalServerError)
             {
-                throw new FlareSolverrException("HTTP StatusCode not 200 or 500. Status is :" + response.StatusCode);
+                throw new FlareSolverrException("HTTP StatusCode not 200 or 500. Status is :" + flaresolverrResponse.StatusCode);
```
mynameisbogdan: Good catch.

src/NzbDrone.Core/IndexerProxies/FlareSolverr/FlareSolverr.cs
```diff
@@ -71,7 +71,20 @@ public override HttpResponse PostResponse(HttpResponse response)
...
+            if (result.Solution.Response.IsNotNullOrWhiteSpace())
+            {
+                return new HttpResponse(
+                    response.Request,
+                    response.Headers,
+                    response.Cookies,
+                    result.Solution.Response,
+                    response.ElapsedTime,
+                    HttpStatusCode.OK);
+            }
```
mynameisbogdan: This would mean always going through FS for bad sites.

Can you provide some more info on the statement `because cf_clearance is bound to the solver's TLS fingerprint` from official sources?

PR #2606, opened 2026-02-18
Fix newznab redirect database migration

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix migration for older versions of SQLite.


mynameisbogdan:
@RetroDadson can you merge this? 

The issue is still there, it was just "fixed" by the bundled sqlite changes which forces newer versions, but platforms where SourceGear.sqlite3 isn't available and the host still uses 3.1x for some reason, the issue is pretty much still there.

RetroDadson (approved):

PR #2625, opened 2026-03-12
New: Configurable per-indexer query timeout

### LINKED ISSUE #586: [FEAT]: Add Indexer Timeout Setting

Not everyone runs Prowlarr under ideal conditions and not all indexers are healthy. Running Prowlarr on low resource hardware in addition to accessing busy indexers during peak time requires the need to adjust timeout values. If coupled with a less than desirable overburdened ISP during peak time only increases the need to adjust timeout values.

A field to set rssfeed and query timeout values when adding or editing indexer options.

We can't always control our environment nor do we control unhealthy indexers. NZBHydra2 has this option and it works beautifully. This is one feature I miss since switching. The only alternative is to use NZBHydra2 which I am attempting to avoid.

### Anything else?

The [Feature Parity with Hydra](https://github.com/Prowlarr/Prowlarr/issues/27) list comments `Timeout No plans to make this an option Q` which I strongly urge you to reconsider the importance of this. This is a requirement for many users.

[AB#1883](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/1883)

simonefil:
## Database Migration
No

## Description
Adds a configurable Query Timeout field (in milliseconds) to each indexer's advanced settings. Default is 100000 ms (100s), matching the previous hardcoded behavior. Valid range: 50 ms to 600000 ms (10 minutes). Improved timeout error messages to show the configured timeout value and suggest adjusting the setting.

Closes https://github.com/Prowlarr/Prowlarr/issues/586

### DIFF
--- a/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
@@ -662,9 +662,9 @@ namespace NzbDrone.Core.Indexers
             request.HttpRequest.SuppressHttpError = true;
             request.HttpRequest.Encoding ??= Encoding;
 
-            if (request.HttpRequest.RequestTimeout == TimeSpan.Zero && Settings.BaseSettings != null)
+            if (request.HttpRequest.RequestTimeout == TimeSpan.Zero && Settings.BaseSettings is { QueryLimit: not null })
             {
-                request.HttpRequest.RequestTimeout = TimeSpan.FromSeconds(Settings.BaseSettings.QueryTimeout);
+                request.HttpRequest.RequestTimeout = TimeSpan.FromSeconds(Settings.BaseSettings.QueryTimeout.Value);
             }
 
             var response = await RetryStrategy
--- a/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
@@ -526,7 +526,8 @@ namespace NzbDrone.Core.Indexers
             catch (TaskCanceledException ex)
             {
                 _indexerStatusService.RecordFailure(Definition.Id);
-                _logger.Warn(ex, "Unable to connect to indexer, possibly due to a timeout. [{0}]", url);
+                var timeoutMs = Settings.BaseSettings?.QueryTimeout ?? 100000;
+                _logger.Warn(ex, "Unable to connect to indexer, possibly due to a timeout ({0} ms). [{1}]", timeoutMs, url);
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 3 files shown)


mynameisbogdan:
50ms is super overkill, the unit should be in seconds as it's more than enough and it makes it easier to configure.

Also all the timeout value logging is not needed IMO.

simonefil:
> 50ms is super overkill, the unit should be in seconds as it's more than enough and it makes it easier to configure.
> 
> Also all the timeout value logging is not needed IMO.

The log message is needed since the logs are the same for HTTP 5XX status codes and timeouts. With the default timeout it was safe to say it was 100% a peoblem of the indexer so it was fine to group them. Since you can lower the timeout it might not be an indexer problem to it's right to differentiate them.

About the timeout. Network delays are always measured in ms, for user friendliness it might be fine to use s as long as you can specify fractions ex: 0.5
I don't agree with 1s minimum, if the indexer is at least in the same country 1s to answer an http request means that their infrastructure is undersized,  if the value is too low it will fail the indexer test and it won't allow you to save which is IMO the correct behaviour.

mynameisbogdan:
> Network delays are always measured in ms

With what is that helping the UX? This is not simply a ConnectionTimeout.

Even good APIs respond in 200-500ms with cached content by Edge, let alone responding under 1000ms with results for a search. Opening a bunch of requests with no purpose other than to fail is not improving anything.

Considering there's also a retry strategy of 3 retries, it's purely bound to annoy a few sites. 

Also think in multitude of users, not only yourself. One site in particular have seen a lot of extra traffic only by adding support for TV searches.

Even 1s is too low IMO.

mynameisbogdan (changes requested):
Null value for default timeout, to make it easier to change the default in the future if needed.

simonefil:
Will this PR ever be merged? Thanks

mynameisbogdan:
> Will this PR ever be merged? Thanks

First step would be to have a working build, right? 😄 

Screenshot

I'm guessing it's going to be eventually, after some polishing.

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -661,6 +662,11 @@ protected virtual async Task<IndexerResponse> FetchIndexerResponse(IndexerReques
             request.HttpRequest.SuppressHttpError = true;
             request.HttpRequest.Encoding ??= Encoding;
 
+            if (request.HttpRequest.RequestTimeout == TimeSpan.Zero && Settings.BaseSettings is { QueryLimit: not null })
```
mynameisbogdan:
```suggestion
            if (request.HttpRequest.RequestTimeout == TimeSpan.Zero && Settings is { BaseSettings.QueryLimit: > 0 })

```

simonefil:
> First step would be to have a working build, right? 😄

This failure appears unrelated to this PR. The build reaches restore and fails on an existing dependency vulnerability. My PR does not touch dependencies or restore/build configuration, so this appears unrelated to the change. Happy to adjust if you prefer handling the dependency update in this PR, though it may be cleaner as a separate fix.

mynameisbogdan:
Latest commit replaces the whole file. Something went wrong there.

simonefil:
> Latest commit replaces the whole file. Something went wrong there.

The only real difference is the edit proposed by you. I don't know why it shows as a whole document diff. Github hallucinations ? 🤨

mynameisbogdan:
https://www.githubstatus.com

Seems so.

mynameisbogdan:
?

You could rebase as it has a mailkit bump.

tylerwiebermusic-max:
Thank you

PR #2650, opened 2026-04-06
Feature: Implement a user controlled rate limit.

### LINKED ISSUE #2635: Prowlarr/Servarr does not have long enough delay between queries and is unintentionally DDOS indexers

## TLDR
Similar issues have been submitted however I am pointing out that this is a major design flaw which turns the servarr into a unintentional DDOS to free indexers. Many queries back to back queries (~50) are not possible as indexers begin to be deactivated and search window begins to dwindle.
### **Explaining design problem with servarr project**
When performing many queries indexers get deactivated because they receive an error due to the indexer timing out prowlarr for attempting to overload the server. This is particularly problematic as upon deactivating the indexer it will be deactivated for an increasing period of time until it is essentially never used. This should really only happen when there are continuous server sided issues (e.g. 500,503,504 ...) as deactivation can only be undone by taking the time to retest the indexers which is hardcoded to 6 hours, which if I am correct has to happen client side for everything that uses prowlarr for queries. 

###  Why is this bad?
So sonarr (or any other adjacent client) will attempt to hit the indexer with as many requests possible between queries. Which regardless of intent when prowlarr follows through without delay looks like a denial of service attack and is clearly attempting to overload the server.

### Even worse Prowlarr is unintentionally DDOS indexers
When a new release occurs or if this project becomes popular enought that many servarr instances querying an indexers at the same time while not abiding by the rate limit will can (possibly) trip cloudflare's ddos protection making that indexer unusable.

If querying the indexers is sequential then when parallel searches is added the 2 above problems will be made much worse.

A queue and delay system should be implemented for each indexer, 

1. A queue so that when multiple clients query the same indexer at the same time they are not sent simultaneously (not sure if prowlarr already has this).
2. A delay that linearly increases for every 429 or "too many requests" (found that some indexers send this instead of 429 which leads to an internal prowlarr error) response that is received from the indexer followed by a cooldown and test period before resuming normal function
3. Add an indexer option to set the delay

example of linear delay function:

$$t=\frac{\Delta t}{n}$$

where $\Delta t$ is a constant span of time and $n$ is the number of requests. And while you could just guess time delays and fit it I think it would be easier and less testing the limits of the server to assume a $\Delta t$ of 60 seconds and $n$ as a positive integer and fit via bisection method by varying n.

Creating a script that measures and estimates the maximum number of requests and then hardcode those values into the code as a lookup table however this is bad if an indexer changes their value.

### Anything else?

I would like to help you add this feature however I am currently busy, unfamiliar with c# and the source code of the project. However I could mock up a more in depth example of this in c# using a functional approach that you can adapt. 
log snippets:

```
2026-03-20 17:19:23.3|Warn|Cardigann|HTTP Error - Res: HTTP/2.0 [GET] https://bitsearch.to/search?q=redacted2&sortBy=created&order=desc&page=1: 429.TooManyRequests (17 bytes)
Too Many Requests
2026-03-20 17:19:23.3|Warn|Cardigann|Request Limit reached for BitSearch. Disabled for 01:00:00

[v2.3.0.5236] NzbDrone.Common.Http.TooManyRequestsException: HTTP request failed: [429:TooManyRequests] [GET] at [https://bitsearch.to/search?q=redacted&sortBy=created&order=desc&page=1]
   at NzbDrone.Core.Indexers.HttpIndexerBase`1.FetchIndexerResponse(IndexerRequest request) in ./NzbDrone.Core/Indexers/HttpIndexerBase.cs:line 696
   at NzbDrone.Core.Indexers.HttpIndexerBase`1.FetchPage(IndexerRequest request, IParseIndexerResponse parser) in ./NzbDrone.Core/Indexers/HttpIndexerBase.cs:line 587
   at NzbDrone.Core.Indexers.HttpIndexerBase`1.FetchReleases(Func`2 pageableRequestChainSelector, SearchCriteriaBase searchCriteria, Boolean isRecent) in ./NzbDrone.Core/Indexers/HttpIndexerBase.cs:line 428
Too Many Requests
```

note  that with bangumi NzbDrone has an exception because 429 error was not returned
```
2026-03-20 17:14:44.2|Warn|Cardigann|https://bangumi.moe/api/v2/torrent/search?query=redacted

[v2.3.0.5236] NzbDrone.Core.Indexers.Exceptions.IndexerException: Error Parsing Rows Selector
   at NzbDrone.Core.Indexers.Definitions.Cardigann.CardigannParser.ParseResponse(IndexerResponse indexerResponse) in ./NzbDrone.Core/Indexers/Definitions/Cardigann/CardigannParser.cs:line 121
   at NzbDrone.Core.Indexers.HttpIndexerBase`1.FetchPage(IndexerRequest request, IParseIndexerResponse parser) in ./NzbDrone.Core/Indexers/HttpIndexerBase.cs:line 591
   at NzbDrone.Core.Indexers.HttpIndexerBase`1.FetchReleases(Func`2 pageableRequestChainSelector, SearchCriteriaBase searchCriteria, Boolean isRecent) in ./NzbDrone.Core/Indexers/HttpIndexerBase.cs:line 428
RequestUri: https://bangumi.moe/api/v2/torrent/search?query=redacted;StatusCode: OK;ContentType: application/json; charset=utf-8;ContentLength: 44;ContentSample: {"success":false,"message":"too frequently"}
```

elio42:
#### Database Migration
NO

#### Description
This pr adds the functionality for the user to change the rate limit for each indexer individually.

##### Behavior

Currently (before the pr) Prowlarr seems to have had a hard-coded lower limit of 2 seconds or the rate limit of the provider if its higher. I implemented this in a way that the user **cannot** set a rate limit lower than the hard-coded limit of 2 seconds, but they **can** set a limit lower that that of the indexer.

As an example lets look at 1337x which currently has a rate limit of 3s:
- The user would be able to set a rate limit of 2 seconds
- But the user cannot set a rate limit lower than 2 seconds

<s>

I implemented it like this because in issue #2635 it was said:

> (...) That said, I'm not against a solution to allow user defined the rate limits per indexer as long they are not lower than our limits.
>
> — @mynameisbogdan

I don't fully understand yet why this would be important. Personally I think that we could allow people to set any rate limit - even 0.

Because even when the rate limit is 0 the requests will happen sequentially, so we will never overwhelm an indexer. And if some indexer doesn't allow a certain cadence of requests I don't see why it makes sense to restrict **ALL** indexers to this 2 second minimum interval. I think that its the users responsibility to choose an adequate rate limit.

I would be glad if someone could comment on why we should not allow lower rate limits than the default - otherwise I would suggest to allow any rate limit.

</s>

#### Todos
- [ ] Tests

* Fixes #2635

### DIFF
--- a/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
@@ -43,7 +43,7 @@ namespace NzbDrone.Core.Indexers
                     { Exception: HttpException { Response.HasHttpServerError: true } } => PredicateResult.True(),
                     _ => PredicateResult.False()
                 },
-                Delay = RateLimit,
+                Delay = GetResolvedRateLimit(),
                 MaxRetryAttempts = 2,
                 BackoffType = DelayBackoffType.Exponential,
                 UseJitter = true,
@@ -80,6 +80,31 @@ namespace NzbDrone.Core.Indexers
         public override IndexerCapabilities Capabilities { get; protected set; }
         public virtual int PageSize => 0;
         public virtual TimeSpan RateLimit => TimeSpan.FromSeconds(2);
+
+        private TimeSpan GetResolvedRateLimit(TimeSpan? requestDelay = null)
+        {
+            var systemDelay = requestDelay ?? TimeSpan.Zero;
+
+            if (systemDelay < RateLimit)
+            {
+                systemDelay = RateLimit;
+            }
+
+            if (Settings?.BaseSettings?.RequestDelay is null or < 2 or > 60)
+            {
+                return systemDelay;
+            }
+
+            var userDelay = TimeSpan.FromSeconds(Settings.BaseSettings.RequestDelay.Value);
+
+            if (userDelay > systemDelay)
+            {
+                _logger.Trace("Using user defined request delay of {0} seconds for indexer {1}", userDelay.TotalSeconds, Definition.Name);
+                return userDelay;
+            }
+
+            return systemDelay;
+        }
 
         public abstract IIndexerRequestGenerator GetRequestGenerator();
         public abstract IParseIndexerResponse GetParser();
@@ -236,11 +261,7 @@ namespace NzbDrone.Core.Indexers
                 return new IndexerDownloadResponse(Encoding.UTF8.GetBytes(request.Url.FullUri));
             }
 
-            if (request.RateLimit < RateLimit)
-            {
-                request.RateLimit = RateLimit;
-            }
-
+            request.RateLimit = GetResolvedRateLimit(request.RateLimit);
             request.AllowAutoRedirect = false;
 
             byte[] fileData;
@@ -636,10 +657,7 @@ namespace NzbDrone.Core.Indexers
         {
             _logger.Debug("Downloading Feed " + request.HttpRequest.ToString(false));
 
-            if (request.HttpRequest.RateLimit < RateLimit)
-            {
-                request.HttpRequest.RateLimit = RateLimit;
-            }
+            request.HttpRequest.RateLimit = GetResolvedRateLimit(request.HttpRequest.RateLimit);
 
             if (_configService.LogIndexerResponse)
             {
@@ -716,11 +734,7 @@ namespace NzbDrone.Core.Indexers
                 request.RequestTimeout = TimeSpan.FromSeconds(15);
             }
 
-            if (request.RateLimit < RateLimit)
-            {
-                request.RateLimit = RateLimit;
-            }
-
+            request.RateLimit = GetResolvedRateLimit(request.RateLimit);
             var response = await _httpClient.ExecuteProxiedAsync(request, Definition);
 
             _eventAggregator.PublishEvent(new IndexerAuthEvent(Definition.Id, !response.HasHttpError, response.ElapsedTime));
--- a/src/NzbDrone.Core/Indexers/IndexerBaseSettings.cs
+++ b/src/NzbDrone.Core/Indexers/IndexerBaseSettings.cs
@@ -16,6 +16,10 @@ namespace NzbDrone.Core.Indexers
                 .GreaterThan(0)
                 .When(c => c.GrabLimit.HasValue)
                 .WithMessage("Should be greater than zero");
+
+            RuleFor(c => c.RequestDelay)
+                .InclusiveBetween(2, 60)
+                .When(c => c.RequestDelay.HasValue);
         }
     }
 
@@ -29,6 +33,9 @@ namespace NzbDrone.Core.Indexers
 
         [FieldDefinition(3, Type = FieldType.Select, Label = "IndexerSettingsLimitsUnit", SelectOptions = typeof(IndexerLimitsUnit), HelpText = "IndexerSettingsLimitsUnitHelpText", Advanced = true)]
         public int LimitsUnit { get; set; } = (int)IndexerLimitsUnit.Day;
+
+        [FieldDefinition(4, Type = FieldType.Number, Label = "IndexerSettingsRequestDelay", HelpText = "IndexerSettingsRequestDelayHelpText", Advanced = true)]
+        public double? RequestDelay { get; set; }
     }
 
     public enum IndexerLimitsUnit
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -428,6 +428,8 @@
   "IndexerSettingsPreferMagnetUrlHelpText": "When enabled, this indexer will prefer the use of magnet URLs for grabs with fallback to torrent links",
   "IndexerSettingsQueryLimit": "Query Limit",
   "IndexerSettingsQueryLimitHelpText": "The number of max queries as specified by the respective unit that {appName} will allow to the site",
+  "IndexerSettingsRequestDelay": "Minimum Request Delay",
+  "IndexerSettingsRequestDelayHelpText": "Minimum delay in seconds between requests for this indexer. The rate limit can only be increased by this setting. Setting a lower value than the default delay will be ignored.",
   "IndexerSettingsRssKey": "RSS Key",
   "IndexerSettingsSeedRatio": "Seed Ratio",
   "IndexerSettingsSeedRatioHelpText": "The ratio a torrent should reach before stopping, empty uses the download client's default. Ratio should be at least 1.0 and follow the indexers rules",


Thefrank:
> I don't fully understand yet why this would be important. Personally I think that we could allow people to set any rate limit - even 0.
> ...
> I think that its the users responsibility to choose an adequate rate limit.

Zero effort DOS. Trackers would blame the app not the user that made the decision that 0 was an awesome number and should be everywhere. Banned app hurts all users of said app.

> Because even when the rate limit is 0 the requests will happen sequentially, so we will never overwhelm an indexer. 

You sure about that? "Never" is a pretty strong statement. One person being stupid might not hurt the tracker but thousands of people being stupid will nuke it pretty hard. DOS -> DDOS via stupid.

elio42:
Fair point regarding the "Fleet impact". I was thinking only about a single instance, but I see how thousands of users adopting a "0 delay" setting could become a considerable problem for trackers.

Anyways, I don't think this would make a big user experience difference anyways.

The current implementation of the pr already respects the 2 second lower limit that existed previously in prowlarr.

ilike2burnthing:
> I implemented this in a way that the user cannot set a rate limit lower than the hard-coded limit of 2 seconds, but they can set a limit lower that that of the indexer.

Without commenting on the merit of the PR as a whole, this should definitely not be how it's done. Hard-coding limits greater than 2s is either because it's required by a site/API, or because without it Prowlarr or Jackett have previously been outright blocked. Allowing users to override this will result in more blocks and/or account bans.

If we're adding specific request delays, it's for a reason. If these need changed or added, that should be reported so it can be done for everyone.

n1ce-0ne:
loving the discussion here, proper checking out all scenarios, tis good, keep at it & im sure it will get sorted.
personally i can see it from all angles, last thing we want is the app being banned by trackers :)

src/NzbDrone.Core/Indexers/IndexerBaseSettings.cs
```diff
@@ -29,6 +29,9 @@ public class IndexerBaseSettings
 
         [FieldDefinition(3, Type = FieldType.Select, Label = "IndexerSettingsLimitsUnit", SelectOptions = typeof(IndexerLimitsUnit), HelpText = "IndexerSettingsLimitsUnitHelpText", Advanced = true)]
         public int LimitsUnit { get; set; } = (int)IndexerLimitsUnit.Day;
+
+        [FieldDefinition(4, Type = FieldType.Number, Label = "IndexerSettingsRequestDelay", HelpText = "IndexerSettingsRequestDelayHelpText", Advanced = true)]
+        public double? RequestDelay { get; set; }
```
mynameisbogdan, line 38: Even as is, this needs some validation to avoid overflows with bad values.

```csharp
            RuleFor(c => c.RequestDelay)
                .InclusiveBetween(2, 60)
                .When(c => c.RequestDelay.HasValue);
```

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -81,6 +82,19 @@ public abstract class HttpIndexerBase<TSettings> : IndexerBase<TSettings>
...
+            var userDelay = Settings?.BaseSettings?.RequestDelay;
+            var delays = new[]
+            {
+                MinimumRateLimit.TotalSeconds,
+                RateLimit.TotalSeconds,
+                userDelay ?? 0,
+                requestDelay?.TotalSeconds ?? 0
+            };
+            return TimeSpan.FromSeconds(delays.Max());
+        }
```
mynameisbogdan, line 107: We should not use a max, but rather check the user value and RateLimit so we can log a message indicating it's using a value set by the user to avoid cases where users asks why it's taking so long when it's their fault.

mynameisbogdan (changes requested):

elio42:
I updated the pr to make sure that the rate limit is not just always more than 2s but _**also**_ more than the rate limit defined in the yaml for that indexer. Tbh, when I started this pr I was not yet quite aware how real the risk of prowlarr being banned by indexers is and that it happened multiple times before.

So the rate limit would now effectively be: `actual_rate_limit = max(user_setting, 2s, rate_limit_from_yaml)`

Furthermore I also implemented the other suggestions.

### Concerning the general merit of the pr

I do agree that if rate limits need to be changed, it should be changed for everyone.

But even so, when a user gets errors back because they are rate limited by an indexer, its not really a good user experience that the user cannot increase the rate limit to get the indexer working again right away. Right now there is no way to get this indexer up and running before it is changed in the yaml definitions.

Furthermore, while I don't know how often this is relevant IRL, there may be situations where some users might need higher rate limits than other users. For example:
1. Users behind a CGNAT or a VPN that share their public IP adress with others.
2. Cloudflare certainly uses "IP Reputation" in their DDOS protection, though I don't know if they use this for their rate limiting directly as well.

### Future idea

I also understand the point:

> "If we're adding specific request delays, it's for a reason. If these need changed or added, that should be reported so it can be done for everyone."

This gave me the idea that maybe in the event of a `429` (too many requests), prowlarr could proactively inform the user about the specific error on the status page with a message along the lines of `It seems that prowlarr sent too many requests to this indexer, please consider reporting this to https://github.com/Prowlarr/Prowlarr/issues.` (Right now, the only way to know what kind of error an indexer had, is to look at the logs)

But I suppose this would be a topic for separate issue.

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -81,6 +82,27 @@ public abstract class HttpIndexerBase<TSettings> : IndexerBase<TSettings>
...
+            var systemDelay = systemDelays.Max();
+
+            if (userDelay.HasValue && userDelay.Value >= systemDelay)
+            {
+                _logger.Trace("Using user defined request delay of {0} seconds for indexer {1}", userDelay.Value, Definition.Name);
+                return TimeSpan.FromSeconds(userDelay.Value);
+            }
+
+            return TimeSpan.FromSeconds(systemDelay);
+        }
```
mynameisbogdan, line 107:
```suggestion
        private TimeSpan ResolvedRateLimit
        {
            get
            {
                if (Settings?.BaseSettings?.RequestDelay is null or < 2 or > 60)
                {
                    return RateLimit;
                }

                var requestDelay = TimeSpan.FromSeconds(Settings.BaseSettings.RequestDelay.Value);

                if (requestDelay > RateLimit)
                {
                    _logger.Trace("Using user defined request delay of {0} seconds for indexer {1}", requestDelay, Definition.Name);

                    return requestDelay;
                }

                return RateLimit;
            }
        }
```

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -29,6 +29,7 @@ public abstract class HttpIndexerBase<TSettings> : IndexerBase<TSettings>
     {
         protected const int MaxNumResultsPerQuery = 1000;
         private const int MaxRedirects = 5;
+        private static readonly TimeSpan MinimumRateLimit = TimeSpan.FromSeconds(2);
```
mynameisbogdan: We already have `public virtual TimeSpan RateLimit => TimeSpan.FromSeconds(2);`

```suggestion
```

mynameisbogdan (changes requested):

elio42:
I went over the PR again, in particular checked if the relevant files are currently subject to any testing and didn't find anything. As such I marked the PR ready for review. Please let me know if I should do anything else.

Niborian:
is there something this is waiting for that I can help with? would be really nice having this😅

Issue #889, opened 2022-03-15
Basic Authentication support when running behind a Reverse Proxy.


Koomongous:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

When running Prowlarr behind a VPN and a Reverse Proxy with Basic Authentication enabled, it cannot communicate with Sonarr/Radarr. This can be solved by making the Reverse Proxy public, or whitelisting the VPN's Public IP within the Reverse Proxy, which is a potential security hazard and not preferred.

### Describe the solution you'd like

I recently saw a feature within the app LunaSea which allows the user to enter the Basic Authentication details (Username and Password) for Sonarr/Radarr and then converts them to base64 encoding. This allows the app to login to the Reverse Proxy and communicate with Sonarrr/Radarr.

I suggest implementing a similar Basic Authentication option under the Advanced Settings for each App.

### Describe alternatives you've considered

Making the Reverse Proxy public, or whitelisting the VPN's Public IP within the Reverse Proxy, which is a potential security hazard and not preferred.

### Anything else?

![Screenshot 2022-03-15 003840](https://user-images.githubusercontent.com/71297607/158283592-61715fb9-24b5-47a9-ab8f-0fa0c006a786.png)

[AB#2792](https://dev.azure.com/Servarr/7ab38f4e-5a57-4d70-84f4-94dd9bc5d6df/_workitems/edit/2792)

bakerboy448:
Generally speaking you should not be going through a reverse proxy to communicate between the \*Arrs anyway - no benefit and only lots of headaches.  Low likelihood of this getting implemented.

The better question is why do you have your api paths requiring double authentication?
The APIs with your setup require both the basic auth and the apikey - there is no need for that and it also does not match the reverse proxy examples in the docs nor SWAG's configs either.

Koomongous:
Due to some weird issue my prowlarr won't work with proxies so I am required to run it through a VPN. By using a Reverse Proxy it can now communicate with my arrs but I use Access Lists to stop just anyone accessing my network, meaning it can only communicate by either entering the Basic Auth or by having it's IP whitelisted. Normally I wouldn't run it like this but I've found that it works.

I realise it's a very edge case problem but it would be nice to have that option one day.

PR #2677, opened 2026-05-21
New: support Basic Auth creds in external app urls

zze0s:
#### Database Migration
NO

#### Description
This PR adds support for Basic Auth creds in URLs for external apps running behind a reverse proxy with Basic Auth enabled.

Previously Prowlarr would fail the validation of the URLs and error when you added Basic Auth creds in the url like: `https://user:pass@host.dev/radarr`.

This is a very basic change/fix. Check if the URL has creds in it, extract them and then set them on the `HttpRequestBuilder` as `NetworkCredential` so they get set as a Basic Auth header.

A more elegant solution would be to add separate user/pass fields and a toggle.

It does not support it for Prowlarr itself, but that can be added.

Tested it with Radarr and Sonarr so far and it works as expected.

Disclosure: I'm no C# expert so this was done with the help of Claude.

#### Todos

* Related #889


mynameisbogdan:
A good idea, but the approach is not user friendly. 

Instead of striping the credentials from the base url, would be better to be added as advanced settings. Similar to what I did for apprise.

zze0s:
I agree that is the better solution :+1: I'll make the changes.

mynameisbogdan:
> It does not support it for Prowlarr itself, but that can be added.

I don't think is a good feature for Prowlarr, as apikey need to remain sole mandatory for the torznab/newznab/grab endpoints and it might encourage some bad setups while they'll decide BA is enough and set the api-key header in the RP. Or some silly stuff like that.

Even regarding this PR, the `/api` should be excluded for basic auth in the RP configuration but since they're synced once every 6 hours there's no harm in double authentication methods I think.

Looks like @bakerboy448 already pointed this out in https://github.com/Prowlarr/Prowlarr/issues/889#issuecomment-1067445804

zze0s:
I see. In this particular case it's Prowlarr running at home and arrs running on a shared seedbox where you can't make any changes to the proxy.

In my own setup I don't have prowlarr behind basic auth, but I'm sure there's plenty of setups where it's used.

bakerboy448:
Only one seedbox companies fail to heed official docs and place api authed points beyond RP refusing to change - USB. Everyone else doesn't auth the API as intended

zze0s:
Add HBD to that list.

mynameisbogdan:
It's okay to use it for the UI if you want to, but right now I'm referring strictly to `/api` usage. 

Sadly basic auth I think it was already remove as a form of authentication on how badly it behaves on Safari, trying to pass multiple credentials in hope the service is fine with one of them. Allows a weird form of brute force.

zze0s:
Changed to use separate user/pass fields and removed the url parsing stuff. Tested with Radarr and Sonarr and still work as intended.

image

src/NzbDrone.Core/Applications/LazyLibrarian/LazyLibrarianSettings.cs
```diff
@@ -44,10 +44,16 @@ public LazyLibrarianSettings()
         [FieldDefinition(1, Label = "LazyLibrarian Server", HelpText = "URL used to connect to LazyLibrarian server, including http(s)://, port, and urlbase if required", Placeholder = "http://localhost:5299")]
         public string BaseUrl { get; set; }
 
-        [FieldDefinition(2, Label = "API Key", Privacy = PrivacyLevel.ApiKey, HelpText = "The ApiKey generated by LazyLibrarian in Settings/Web Interface")]
+        [FieldDefinition(2, Label = "ApplicationAuthUsername", HelpText = "ApplicationAuthUsernameHelpText", Privacy = PrivacyLevel.UserName, Advanced = true)]
+        public string AuthUsername { get; set; }
+
+        [FieldDefinition(3, Label = "ApplicationAuthPassword", HelpText = "ApplicationAuthPasswordHelpText", Type = FieldType.Password, Privacy = PrivacyLevel.Password, Advanced = true)]
+        public string AuthPassword { get; set; }
```
mynameisbogdan: Can you put them after ApiKey?

mynameisbogdan (changes requested):

src/NzbDrone.Core/Applications/LazyLibrarian/LazyLibrarianSettings.cs
zze0s: Ofc! Done!

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -45,6 +45,10 @@
   "AppUpdated": "{appName} Updated",
   "AppUpdatedVersion": "{appName} has been updated to version `{version}`, in order to get the latest changes you'll need to reload {appName}",
   "Application": "Application",
+  "ApplicationAuthPassword": "Auth Password",
+  "ApplicationAuthPasswordHelpText": "HTTP Basic Auth password used by Prowlarr when accessing this application. Set if the application sits behind a Basic-auth-protected reverse proxy.",
+  "ApplicationAuthUsername": "Auth Username",
+  "ApplicationAuthUsernameHelpText": "HTTP Basic Auth username used by Prowlarr when accessing this application. Set if the application sits behind a Basic-auth-protected reverse proxy.",
```
mynameisbogdan:
```suggestion
  "ApplicationAuthPassword": "Auth Password",
  "ApplicationAuthPasswordHelpText": "HTTP Basic Auth password used by {appName} when accessing this application. Set if the application sits behind a Basic-auth-protected reverse proxy.",
  "ApplicationAuthUsername": "Auth Username",
  "ApplicationAuthUsernameHelpText": "HTTP Basic Auth username used by {appName} when accessing this application. Set if the application sits behind a Basic-auth-protected reverse proxy.",
```

mynameisbogdan (changes requested):
Otherwise, all looks reasonable to me.

zze0s:
Great :partying_face: I appreciate the reviews @mynameisbogdan!

mynameisbogdan (approved):
Suggestion for squash title: `New: Basic Auth support for sync applications`

augustuen (approved):

PR #2696, opened 2026-06-10
fix: Cloudflare detection without requiring Server header

dnl1:
## Problem

The `IsCloudflareProtected` method in `CloudFlareDetectionService` required a `Server: cloudflare` (or `cloudflare-nginx`, `ddos-guard`) response header **before** it would check the response content for Cloudflare/DDoS-Guard patterns. 

Many Cloudflare-protected sites do **not** include this header, causing the early return at line 23-26 to skip all content analysis and return `false` — preventing FlareSolverr from being activated. The request then fails with a direct HTTP error (403, 302 redirect to cdn-cgi challenge), and the indexer gets disabled.

## Affected Indexers

- **TorrentGalaxyClone**: Returns HTTP 302 redirect to `cdn-cgi/content?id=...` (Cloudflare redirect challenge) — `Server` header missing
- **kickasstorrents.to**: Returns HTTP 403 with Cloudflare HTML body — `Server` header missing on many mirrors
- **EZTV** (and potentially others): Same 403 pattern, same missing `Server` header

FlareSolverr itself can successfully bypass these sites when called directly (tested manually).

## Root Cause

```csharp
// OLD CODE - checks Server header FIRST, returns false immediately if missing
if (!response.Headers.Any(i => i.Key != null && 
    i.Key.ToLower() == "server" && 
    CloudflareServerNames.Contains(i.Value.ToLower())))
{
    return false;  // ← Never reaches content analysis below
}
```

## Fix

1. **Content analysis moved first** — Status code + HTML title/body patterns are checked regardless of `Server` header presence
2. **`cdn-cgi` detection in 403/503 bodies** — Additional pattern match for Cloudflare responses without standard titles
3. **Redirect challenge detection** — New handler for HTTP 301/302/303/307 responses with `cdn-cgi` in the `Location` header or response body (TorrentGalaxyClone case)
4. **Server header kept as additional signal** — Not removed, just no longer a gate that blocks content analysis

### DIFF
--- a/src/NzbDrone.Core/Http/CloudFlare/CloudFlareDetectionService.cs
+++ b/src/NzbDrone.Core/Http/CloudFlare/CloudFlareDetectionService.cs
@@ -20,12 +20,7 @@ namespace NzbDrone.Core.Http.CloudFlare
 
         public static bool IsCloudflareProtected(HttpResponse response)
         {
-            if (!response.Headers.Any(i => i.Key != null && i.Key.ToLower() == "server" && CloudflareServerNames.Contains(i.Value.ToLower())))
-            {
-                return false;
-            }
-
-            // detect CloudFlare and DDoS-GUARD
+            // detect CloudFlare and DDoS-GUARD via content analysis (most reliable, check first)
             if (response.StatusCode.Equals(HttpStatusCode.ServiceUnavailable) ||
                 response.StatusCode.Equals(HttpStatusCode.Forbidden))
             {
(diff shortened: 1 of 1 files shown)


ghostkernel-core:
it seems like this solves the new header removal of flaresolvarrv2 issues for many trackers. Can anyone review and merge it for a release please?

mynameisbogdan:
This class follows the changes from the FlareSolverrSharp library at <https://github.com/FlareSolverr/FlareSolverrSharp/blob/master/src/FlareSolverrSharp/ChallengeDetector.cs>.

These would make sense with actual examples and actual linking to cloudflare documentation if possible, and not LLM induced problem and fixing that sometimes are just hallucinations that provide low or even negative quality.

Considering the `many Cloudflare-protected sites do not include this header` statement, I believe the operator should be able to provide examples.

> 3\. **Redirect challenge detection** — New handler for HTTP 301/302/303/307 responses with `cdn-cgi` in the `Location` header or response body (TorrentGalaxyClone case)

I don't think that's a cloudflare challenge, but their own. I might wrong, but I'm unable to trigger it even from VPN IPs.

Most popular protected CF site:
```bash
curl -sI https://1337x.to/ | grep -i '^server:'

server: cloudflare
```

dnl1:
Closing in favor of a more targeted approach. Testing showed all affected sites do return `server: cloudflare`, so removing that gate isn't justified. Will open a new PR focused on adding `cf-mitigated: challenge` header detection instead.

PR #2701, opened 2026-06-11
Fixed: (RuTracker) Word boundary in RUS token duplicate check

hartmanfrost:
#### Database Migration

NO

#### Description

The "Add RUS to titles" option (#1434) guards against appending a duplicate token with:

```csharp
!Regex.Match(title, "\bRUS\b", RegexOptions.IgnoreCase).Success
```

In a non-verbatim C# string `\b` is the backspace character (U+0008), not a regex word boundary, so the pattern can never match and ` RUS` is appended even when the title already contains `RUS` (e.g. releases that already carry a `RUS` language tag end up as `... RUS RUS`). One-character fix to a verbatim string literal, plus a regression test.

#### Todos

- [x] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/RuTrackerTests/RuTrackerTitleParserFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/RuTrackerTests/RuTrackerTitleParserFixture.cs
@@ -1,0 +1,24 @@
+using System.Collections.Generic;
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Indexers;
+using NzbDrone.Core.Indexers.Definitions;
+
+namespace NzbDrone.Core.Test.IndexerTests.RuTrackerTests
+{
+    [TestFixture]
+    public class RuTrackerTitleParserFixture
+    {
+        private static readonly ICollection<IndexerCategory> TvCategories = new List<IndexerCategory> { NewznabStandardCategory.TVHD };
+
+        private readonly RuTrackerTitleParser _titleParser = new();
+
+        [TestCase("Series Title (2019) WEB-DL 1080p", "Series Title (2019) WEB-DL 1080p RUS")]
+        [TestCase("Series Title (2019) WEB-DL 1080p RUS", "Series Title (2019) WEB-DL 1080p RUS")]
+        public void should_add_rus_to_title_without_duplicating(string title, string expected)
+        {
+            _titleParser.Parse(title, TvCategories, stripCyrillicLetters: false, addRussianToTitle: true)
+                .Should().Be(expected);
+        }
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core.Test/IndexerTests/RuTrackerTests/RuTrackerTitleParserFixture.cs
```diff
@@ -0,0 +1,30 @@
...
+            _titleParser.Parse("The Irishman (2019) WEB-DL 1080p", TvCategories, stripCyrillicLetters: false, addRussianToTitle: true)
+                .Should().Be("The Irishman (2019) WEB-DL 1080p RUS");
+        }
+
+        [Test]
+        public void should_not_duplicate_rus_in_title()
+        {
+            _titleParser.Parse("The Irishman (2019) WEB-DL 1080p RUS", TvCategories, stripCyrillicLetters: false, addRussianToTitle: true)
+                .Should().Be("The Irishman (2019) WEB-DL 1080p RUS");
+        }
```
mynameisbogdan: This should be combined into a single test with multiple test cases, not one test per case.

And replace the titles with `Movie Title` or `Series Title`.

mynameisbogdan (changes requested):

src/NzbDrone.Core.Test/IndexerTests/RuTrackerTests/RuTrackerTitleParserFixture.cs
hartmanfrost: Done — combined into a single parameterized test and replaced the titles with a generic `Series Title` placeholder.

src/NzbDrone.Core/Indexers/Definitions/RuTracker.cs
```diff
@@ -1769,7 +1769,7 @@ public string Parse(string title,
             }
 
             // language fix: all rutracker releases contains russian track
-            if (addRussianToTitle && (IsAnyTvCategory(categories) || IsAnyMovieCategory(categories)) && !Regex.Match(title, "\bRUS\b", RegexOptions.IgnoreCase).Success)
+            if (addRussianToTitle && (IsAnyTvCategory(categories) || IsAnyMovieCategory(categories)) && !Regex.Match(title, @"\bRUS\b", RegexOptions.IgnoreCase).Success)
```
mynameisbogdan, line 1772:
```suggestion
            if (addRussianToTitle && (IsAnyTvCategory(categories) || IsAnyMovieCategory(categories)) && !Regex.IsMatch(title, @"\bRUS\b", RegexOptions.IgnoreCase))
```

PR #2708, opened 2026-06-18
Fix: Cardigann dateparse assumes local TZ for timezone-less timestamps

bughaver:
#### Database Migration

NO

#### Description

When a Cardigann YML definition uses the `dateparse` filter with a timezone-less format (e.g. `yyyy-MM-dd HH:mm:ss`), `ParseDateTimeGoLang` returns a `DateTime` with `DateTimeKind.Unspecified`. The filter chain then formats this with `Rfc1123ZPattern` which includes the `z` (UTC offset) specifier. For `Unspecified` `DateTime` values, .NET treats them as `Local` and uses the **system/container timezone offset**, corrupting dates whenever the container TZ differs from UTC.

**Concrete example:** TorrentLeech sends `addedTimestamp` in UTC as `2026-06-03 01:01:21` (format: `yyyy-MM-dd HH:mm:ss`, no timezone). With a non-UTC container TZ (e.g. `US/Pacific`, UTC-7), the `dateparse` filter produces a shifted offset in the output string, making the effective UTC time off by the container's UTC offset. This causes downstream consumers (e.g. Sonarr delay profiles) to see an incorrect release age.

**Root Cause:** `ParseDateTimeGoLang` uses `DateTimeStyles.None` for both its fast path (`TryParseExact`) and slow path (`ParseExact`). When the format lacks timezone specifiers, this produces `DateTimeKind.Unspecified`.

**Fix:** After parsing, if the resulting `DateTime` has `DateTimeKind.Unspecified`, assume UTC via `DateTime.SpecifyKind`. This aligns with the existing codebase convention — for example `AnimeBytes.cs:388` uses `DateTimeStyles.AssumeUniversal` for the same timezone-less format:

```csharp
torrent.UploadTime = DateTime.ParseExact(row.upload_time, "yyyy-MM-dd HH:mm:ss",
    CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal);
```

Timezone-aware formats (`-07:00`, `Z07:00`, `zzz`, etc.) already produce `Local` or `Utc` `DateTimeKind` and are unaffected.

#### Todos

- [x] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/DateTimeUtilFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/DateTimeUtilFixture.cs
@@ -39,5 +39,36 @@ namespace NzbDrone.Core.Test.ParserTests
                 .ToString(standardFormat, CultureInfo.InvariantCulture)
                 .Should().Be(expectedDate);
         }
+
+        [TestCase("2026-06-03 01:01:21", "yyyy-MM-dd HH:mm:ss")]
+        [TestCase("2022-08-08 02:07:39", "yyyy-MM-dd HH:mm:ss")]
+        [TestCase("2022-01-15 12:30:45", "yyyy-MM-dd HH:mm:ss")]
+        public void parse_datetime_golang_timezone_less_should_assume_utc(string dateInput, string format)
+        {
+            var result = DateTimeUtil.ParseDateTimeGoLang(dateInput, format);
+
+            result.Kind.Should().Be(DateTimeKind.Utc);
+            result.Year.Should().Be(int.Parse(dateInput.AsSpan(0, 4)));
+            result.Month.Should().Be(int.Parse(dateInput.AsSpan(5, 2)));
+            result.Day.Should().Be(int.Parse(dateInput.AsSpan(8, 2)));
+        }
+
+        [TestCase("2026-06-03 01:01:21", "2006-01-02 15:04:05")]
+        [TestCase("2022-08-08 02:07:39", "2006-01-02 15:04:05")]
+        public void parse_datetime_golang_golang_format_timezone_less_should_assume_utc(string dateInput, string format)
+        {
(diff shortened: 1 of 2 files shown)


mynameisbogdan:
> **Concrete example:** TorrentLeech sends `addedTimestamp` in UTC as `2026-06-03 01:01:21` (format: `yyyy-MM-dd HH:mm:ss`, no timezone).

TL is using the `Site Timezone` from your profile, that' why it's `local` for this indexer. 

The description seems to be LLM hallucinations (problem, root cause and fix), based on the wrong premise.

bughaver:
@mynameisbogdan  my account timezone is set to UTC, so the addedTimestamp from TL was in UTC when this was observed. The timezone offset in the parsed date came from the container TZ (non-UTC), not from the TL profile setting.

mynameisbogdan:
You're supposed to use your local time, as users will use local when browsing the site.

augustuen:
Closing due to LLM hallucinations. 

The parser doesn't assume it's UTC, it assumes it's local time, because some indexers (like torrentleech) have a setting to set your local timezone. This is also obviously from the Cardigann definition for torrentleech which specifically mentions that it uses your profile's timezone.

mynameisbogdan:
Tried the changes locally and unless everyone sets after this change the `Site Timezone` to UTC, they'll get wrong publish dates. 

Basically fix for one user, break for everyone else.

FYI also all this logic could be replaced with `DateTimeStyles.AssumeUniversal`.

bughaver:
Ok thanks, before I raised the pr I confirmed with TL over irc that dateAdded was utc but maybe they weren't aware its profile time. Sorry for the confusion

mynameisbogdan:
Maybe they thought you're asking about the IRC announces? 

You'll find the option in Profile > Edit.

PR #2714, opened 2026-06-24
[Toloka] Add "Add UKR to Title" option

sigito:
#### Database Migration
NO

#### Description
Adds an optional "Add UKR to Title" setting to the Toloka indexer. When enabled, appends `UKR` to the end of all release titles to improve language detection by Sonarr and Radarr. Specifically it addresses two potential issues:
1. Not picking up Ukrainian language for the releases that do not explicitly mention the language.
2. Treating entries from Toloka as English, since it is default if none specified.

The change mirrors an equivalent `addukrainiantotitle` option already available in Cardigann-based indexers such as [Utopia](https://github.com/Prowlarr/Indexers/blob/master/definitions/v11/utopia.yml#L37-L39).

#### Todos
- [x] Tests

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/TolokaTests/TolokaTitleParserFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/TolokaTests/TolokaTitleParserFixture.cs
@@ -41,5 +41,14 @@ namespace NzbDrone.Core.Test.IndexerTests.TolokaTests
             result.Should().EndWith("UKR");
             result.Should().Contain("Castlevania");
         }
+
+        [Test]
+        public void should_not_append_ukr_when_already_present()
+        {
+            var result = _parser.Parse("Castlevania S03E01 1080p WEB-DL UKR", new List<IndexerCategory>(), addUkrainianToTitle: true);
+
+            result.Should().EndWith("UKR");
+            result.Should().NotContain("UKR UKR");
+        }
     }
 }
--- a/src/NzbDrone.Core.Test/IndexerTests/TolokaTests/TolokaTitleParserFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/TolokaTests/TolokaTitleParserFixture.cs
@@ -17,38 +17,14 @@ namespace NzbDrone.Core.Test.IndexerTests.TolokaTests
             _parser = new TolokaTitleParser();
         }
 
-        [Test]
-        public void should_not_append_ukr_when_disabled()
+        [TestCase("Castlevania S03E01 1080p WEB-DL", false, "Castlevania S03E01 1080p WEB-DL")]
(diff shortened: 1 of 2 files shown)


mynameisbogdan:
Needs to be added only if not already in the title.

See #2701 since it has a word boundary fix on how to add only if needed, and a single test with multiple test cases added.

src/NzbDrone.Core.Test/IndexerTests/TolokaTests/TolokaTitleParserFixture.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        [Test]
+        public void should_append_ukr_after_cyrillic_stripping()
+        {
+            var result = _parser.Parse("Замок Кастлеванія / Castlevania S03E01 1080p WEB-DL", new List<IndexerCategory>(), stripCyrillicLetters: true, addUkrainianToTitle: true);
+
+            result.Should().EndWith("UKR");
+            result.Should().Contain("Castlevania");
+        }
```
mynameisbogdan: This can be added a single test with multiple test cases.

sigito: Done. Also remove the test case for cyrillic stripping, as it is not relevant.

src/NzbDrone.Core/Indexers/Definitions/Toloka.cs
```diff
@@ -511,6 +511,11 @@ public string Parse(string title, ICollection<IndexerCategory> categories, bool
             // replace multiple spaces with a single space
             title = Regex.Replace(title, @"\s+", " ");
 
+            if (addUkrainianToTitle && !Regex.IsMatch(title, @"\bUKR\b", RegexOptions.IgnoreCase))
```
mynameisbogdan: Can you do it like Rutracker, for tv and movies only too?

sigito: Ah, yeah, sure. Thought you only meant the regex check.

mynameisbogdan: My bad, indeed I didn't even noticed till I checked the rutracker again and realized that this might affect other stuff like music, books, etc. too.

sigito: Not sure I got it. Should I add the TV/Movie check or no?

mynameisbogdan: Yes, you should add the TV/Movie check.

sigito: Should be all ready now.

Issue #2684, opened 2026-05-26
Prowlarr not getting the best results due to sorting by date


bolshoytoster:
### I attest that there is not an existing issue for this?

- [x] I have searched the existing open and closed issues

### I attest this is not related to a Cardigann YML Indexer.

- [x] I confirm this is not related to a Cardigann YML Indexer

### Current Behavior

Prowlarr's definition for Knaben.org when searching [always sorts by newest, with a limit of 100 results](https://github.com/Prowlarr/Prowlarr/blob/cb80af4be1dfe277b4b1a71f026683a738cfe50e/src/NzbDrone.Core/Indexers/Definitions/Knaben.cs#L207-L210), which means it sometimes misses the best releases for some shows (tested with Lucifer (2016)).

This is partly because when you sort by date, the results are usually filled with lots of torrents with 0 seeders, which aren't very useful.

From a cursory glance, it looks like some other indexer definitions do this as well, so it may be an intentional decision.

### Expected Behavior

Prowlarr should be able to pick up as many good torrents as possible, for example, when searching for 'Lucifer S01' on Knaben, it should be able to pick up 'Lucifer S01 1080p BluRay HEVC DD5.1 x265'.

It can do this by either changing to sort by seeders when searching (obviously not with an empty query), making the limit for number of results >100 (doesn't completely solve the issue, and results in more traffic/load on the indexer), or by paging the results to get them all (still more traffic/load).

### Steps To Reproduce

1. Add the Knaben.org indexer
2. Search for 'Lucifer S01'
3. Notice 'Lucifer S01 1080p BluRay HEVC DD5.1 x265' is missing, even though it's the 3rd highest seeded result

### Environment

```markdown
- OS: Linux archlinux 7.0.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 17 May 2026 17:23:07 +0000 x86_64 GNU/Linux
- Prowlarr: 2.3.5.5327
- Docker Install: No
- Using Reverse Proxy: No
- Browser:
```

### What branch are you running?

Master

### Trace Logs?

There are no relevant logs.

### I attest that Trace Logs have been provided as applicable. Reports will be closed if the required logs are not provided.

- [x] I attest that I have read and followed the steps in the wiki link above and provided the required trace logs - the logs contain `trace` - that are relevant and show this issue.

mynameisbogdan:
[...]

PR #2716, opened 2026-06-27
New: (IPTorrents) Add Sort by Seeders option

johntmorehead:
#### Database Migration
NO

#### Description
Adds a **Sort By** option to the IPTorrents indexer — a dropdown (`Default` / `Seeders`). When set to **Seeders**, searches request results ordered by seeders (`o=seeders`) instead of the site default (date).

IPTorrents returns a limited page of results, so for queries with many matches the best-seeded releases can fall outside the returned page and never reach the apps. This lets users opt into seeder-ordered results so the most-available releases surface first. The RSS/recent feed is always left in the site's default order so new releases are still picked up. It mirrors the existing `FreeLeechOnly` setting pattern.

The default is **Default** (site order — retains current behavior).

The new "Sort By" dropdown on the IPTorrents indexer settings (directly below "Freeleech Only"):

Sort By setting

Expanded, showing the options:

Sort By options

#### Todos
- [x] Tests

* Related to #2684


mynameisbogdan:
And this is going to mess with RSS syncs.

johntmorehead:
> And this is going to mess with RSS syncs.

Oh man, I didn't even think of that, good catch, thank you. I added a check to make sure that the term and imdbID are empty, so the RSS feed should return date sorted now. Also added a test to prove it works.

src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
```diff
@@ -199,6 +199,13 @@ private IEnumerable<IndexerRequest> GetPagedRequests(string term, SearchCriteria
                 qc.Add("free", "on");
             }
 
+            // Only sort actual searches by seeders, never the RSS/recent feed, which must stay
+            // in the site's default (date) order so new releases are still seen.
```
mynameisbogdan: This comment provides no value to the user, just the LLMs. 

This information would be better to be put inside the settings section for the new option, for users to know what's doing.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -437,6 +438,7 @@
   "IndexerSettingsSeedRatioHelpText": "The ratio a torrent should reach before stopping, empty uses the download client's default. Ratio should be at least 1.0 and follow the indexers rules",
   "IndexerSettingsSeedTime": "Seed Time",
   "IndexerSettingsSeedTimeHelpText": "The time a torrent should be seeded before stopping, empty uses the download client's default",
+  "IndexerSettingsSortBySeeders": "Sort by Seeders",
```
mynameisbogdan: One huge drawback to this being boolean, it's going to require a db migration if someone will ask for another kind of sorting. 😂 

```suggestion
  "IndexerSettingsSearchSortBySeeders": "Search sorting by Seeders",
```

Maybe indicate it's for search only would be more practical.

johntmorehead: Yeah that's fair, but I feel like this is a pretty high value sort, its definitely the first thing I and anyone else I know does most the time. That being said, if you think the tradeoff creates more issues, I get it.

src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
johntmorehead: Probably makes sense to have a comment in both places: the one at the setting explaining that this only affects live searches not RSS, and then the comment here showing that functionality is being implemented correctly. Or is that just overkill? Also, fair to update the text in the UI as well, so people don't get confused or worried?

johntmorehead: I realize my last question was answered by the suggestion you made in the settings file, nvm.

src/NzbDrone.Core/Localization/Core/en.json
johntmorehead: Also, the more I think of it, I could just switch it to a drop-down box or something with a "sort by" and the choice of "seeder" or something else right? Seems like a more future-proof design and achieves the same thing?

mynameisbogdan: Yea, you could add something like `Default = 0`, `Seeders = 1` in an enum.

johntmorehead: I made the change to an enum/dropdown, any other thoughts?

src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
```diff
@@ -199,6 +199,11 @@ private IEnumerable<IndexerRequest> GetPagedRequests(string term, SearchCriteria
                 qc.Add("free", "on");
             }
 
+            if (Settings.Sort == (int)IPTorrentsSort.Seeders && !searchCriteria.IsRssSearch)
```
mynameisbogdan:
```suggestion
            if (!searchCriteria.IsRssSearch && Settings.SearchSortBy == (int)IPTorrentsSort.Seeders)
```

src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
```diff
@@ -432,9 +437,21 @@ public class IPTorrentsSettings : CookieTorrentBaseSettings
         [FieldDefinition(4, Label = "IndexerSettingsFreeleechOnly", Type = FieldType.Checkbox, HelpText = "IndexerIPTorrentsSettingsFreeleechOnlyHelpText")]
         public bool FreeLeechOnly { get; set; }
 
+        [FieldDefinition(5, Label = "IndexerSettingsSearchSortBy", Type = FieldType.Select, SelectOptions = typeof(IPTorrentsSort), HelpText = "IndexerIPTorrentsSettingsSortHelpText")]
+        public int Sort { get; set; }
```
mynameisbogdan:
```suggestion
        [FieldDefinition(5, Label = "IndexerIPTorrentsSettingsSearchSortBy", Type = FieldType.Select, SelectOptions = typeof(IPTorrentsSort), HelpText = "IndexerIPTorrentsSettingsSearchSortByHelpText")]
        public int SearchSortBy { get; set; }
```

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -433,6 +434,7 @@
   "IndexerSettingsQueryLimit": "Query Limit",
   "IndexerSettingsQueryLimitHelpText": "The number of max queries as specified by the respective unit that {appName} will allow to the site",
   "IndexerSettingsRssKey": "RSS Key",
+  "IndexerSettingsSearchSortBy": "Sort By",
```
mynameisbogdan:
```suggestion
```

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -381,6 +381,7 @@
   "IndexerIPTorrentsSettingsCookieUserAgent": "Cookie User-Agent",
   "IndexerIPTorrentsSettingsCookieUserAgentHelpText": "User-Agent associated with cookie used from Browser",
   "IndexerIPTorrentsSettingsFreeleechOnlyHelpText": "Search freeleech releases only",
+  "IndexerIPTorrentsSettingsSortHelpText": "Select the order in which to sort search results. The RSS/recent feed is always left in the site's default order, so new releases are still picked up.",
```
mynameisbogdan:
```suggestion
  "IndexerIPTorrentsSettingsSearchSortBy": "Search Sort By",
  "IndexerIPTorrentsSettingsSearchSortByHelpText": "Select the order in which to sort search results",
```

mynameisbogdan (changes requested):

src/NzbDrone.Core/Localization/Core/en.json
mynameisbogdan: Yea, see last suggestions that are mostly naming related.

johntmorehead: Yeah I like all of those, totally make sense. I batch accepted them and committed. Looks like some of the checks were not successful, but I assume those are not related to the naming changes, but am I missing something?

src/NzbDrone.Core.Test/IndexerTests/IPTorrentsTests/IPTorrentsRequestGeneratorFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var page = results.GetAllTiers().First().First();
+
+            page.Url.FullUri.Should().NotContain("o=seeders");
+        }
+
+        [Test]
+        public void should_sort_by_seeders_when_enabled()
+        {
+            Subject.Settings.Sort = (int)IPTorrentsSort.Seeders;
```
mynameisbogdan: This needs updating too.

mynameisbogdan (changes requested):

src/NzbDrone.Core.Test/IndexerTests/IPTorrentsTests/IPTorrentsRequestGeneratorFixture.cs
johntmorehead: Yeah, duh, I should have thought about that, sorry. Should be fixed now.

johntmorehead:
Is anything further needed from me on this? Happy to make any changes necessary, and if it just needs to be reviewed by the right people I understand, just wanted to check in.

augustuen (approved):

PR #2734, opened 2026-07-15
Fixed: (IPTorrents) Advertise correct page size in capabilities

spinningrustcc:
IPTorrents serves 50 results per page, but the indexer did not override LimitsDefault/LimitsMax, inheriting the generic default of 100 from the IndexerCapabilities constructor.

The advertised limit must match the site's real page size, because IPTorrentsRequestGenerator derives the page number from it:

    var page = (int)(searchCriteria.Offset / searchCriteria.Limit) + 1;
    qc.Add("p", page.ToString());

Advertising 100 against a 50-row page therefore breaks paging for apps that page by the advertised limit. Sonarr computes its page size from these caps (Math.Min(100, Math.Max(DefaultPageSize, MaxPageSize)) = 100), requests limit=100, receives 50, and its IsFullPage(50 >= 100) check is false — so it treats the short page as the last page and stops after page one. Everything past the first 50 results becomes unreachable, even though offset paging works correctly.

Declaring 50 makes IsFullPage true at a full 50-row page, so paging continues via the existing offset -> p translation. No behaviour change for callers that do not page.

Matches the existing pattern in Redacted, AlphaRatio, AvistaZ and others which already declare their real page size.

#### Database Migration
YES - XXXX | NO

#### Description
A few sentences describing the overall goals of the pull request's commits.

#### Todos
- [ ] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
@@ -84,6 +84,8 @@ namespace NzbDrone.Core.Indexers.Definitions
         {
             var caps = new IndexerCapabilities
             {
+                LimitsDefault = 50,
+                LimitsMax = 50,
                 TvSearchParams = new List<TvSearchParam>
                 {
                     TvSearchParam.Q, TvSearchParam.Season, TvSearchParam.Ep, TvSearchParam.ImdbId


mynameisbogdan:
Claude, this can be set in your operator's profile. 

Screenshot

And advertising anything lower than 100 will make the other apps to paginate even if number of results on a page are between 50 and 100.

spinningrustcc:
Thanks for pointing that out — I'd looked for that setting and missed it, and it
does fix things. Setting it to 100 took searches from 50 to 100 results per page,
and Sonarr's season search went from 50 to 136 releases, since it now pages
instead of stopping after page one.

That also means this PR is wrong as written, and I should say so plainly. If the
page size is user-configurable, hardcoding `LimitsMax = 50` would break anyone
who has set 100: Sonarr computes `PageSize = 50`, requests `offset=50`, and the
generator's `page = (offset / limit) + 1` asks for page 2 — which at 100/page is
rows 101–200, silently skipping 51–100. That's worse than the current behaviour.

The bit I'm still unsure about: the profile field defaults to 0 (= default = 50),
so out of the box Prowlarr advertises 100 while the site serves 50, and callers
that page by the advertised limit stop after the first page. Since
`SupportsPagination` only gates offset handling and Prowlarr doesn't aggregate
pages internally, the advertised limit has to match what the site actually
serves — and with that being per-user, I don't think any constant is right here.

So before I put more into this, which direction would you prefer, if any?

1. Expose the page size as an indexer setting (defaulting to 50) so the caps
   reflect the user's profile.
2. If IPTorrents accepts a per-page parameter on the request, set it there and
   keep advertising 100 — deterministic, no setting needed. I couldn't find one,
   but you'd know better.
3. Treat the profile setting as the answer and close this — no objection from me.

Happy to implement 1 or 2 if either is welcome. Thanks again for the pointer.

mynameisbogdan:
3 is the sane option, like I said already anything less than 100 will trigger more requests due to arrs thinking pagination is needed since `IsFullPage()` is going to be `true`.

PR #2753, opened 2026-07-25
Fixed: (Indexer) Update GGN Nintendo Categories

chickenhuntr:
#### Database Migration
NO

#### Description
Make querying for Nintendo Games on GGN nicer via Prowlarr.

Previously had to use Console/Other which would produce a lot of bad results, now each console has it's own category making finding specific games eaiser.

n/a 

#### Todos
- [X] Tests (Manually checked the categories in search) 

n/a

### DIFF
--- a/src/NzbDrone.Core/Indexers/NewznabStandardCategory.cs
+++ b/src/NzbDrone.Core/Indexers/NewznabStandardCategory.cs
@@ -24,6 +24,17 @@ namespace NzbDrone.Core.Indexers
         public static readonly IndexerCategory ConsoleWiiU = new(1130, "Console/WiiU");
         public static readonly IndexerCategory ConsoleXBoxOne = new(1140, "Console/XBox One");
         public static readonly IndexerCategory ConsolePS4 = new(1180, "Console/PS4");
+        public static readonly IndexerCategory ConsoleGameBoy = new(1190, "Console/Game Boy");
+        public static readonly IndexerCategory ConsoleGameBoyAdvance = new(1200, "Console/Game Boy Advance");
+        public static readonly IndexerCategory ConsoleGameBoyColor = new(1210, "Console/Game Boy Color");
+        public static readonly IndexerCategory ConsoleNES = new(1220, "Console/NES");
+        public static readonly IndexerCategory ConsoleNintendo64 = new(1230, "Console/Nintendo 64");
+        public static readonly IndexerCategory ConsoleGameCube = new(1240, "Console/Nintendo GameCube");
+        public static readonly IndexerCategory ConsolePokemonMini = new(1250, "Console/Pokemon Mini");
+        public static readonly IndexerCategory ConsoleSNES = new(1260, "Console/SNES");
+        public static readonly IndexerCategory ConsoleSwitch = new(1270, "Console/Switch");
+        public static readonly IndexerCategory ConsoleVirtualBoy = new(1280, "Console/Virtual Boy");
+        public static readonly IndexerCategory ConsoleNewNintendo3DS = new(1290, "Console/New Nintendo 3DS");
 
         public static readonly IndexerCategory Movies = new(2000, "Movies");
         public static readonly IndexerCategory MoviesForeign = new(2010, "Movies/Foreign");
@@ -120,6 +131,17 @@ namespace NzbDrone.Core.Indexers
             ConsoleWiiU,
             ConsoleXBoxOne,
             ConsolePS4,
+            ConsoleGameBoy,
+            ConsoleGameBoyAdvance,
+            ConsoleGameBoyColor,
+            ConsoleNES,
+            ConsoleNintendo64,
+            ConsoleGameCube,
+            ConsolePokemonMini,
+            ConsoleSNES,
+            ConsoleSwitch,
+            ConsoleVirtualBoy,
+            ConsoleNewNintendo3DS,
             Movies,
             MoviesForeign,
             MoviesOther,
@@ -202,7 +224,18 @@ namespace NzbDrone.Core.Indexers
                     ConsolePSVita,
                     ConsoleWiiU,
                     ConsoleXBoxOne,
-                    ConsolePS4
+                    ConsolePS4,
+                    ConsoleGameBoy,
+                    ConsoleGameBoyAdvance,
+                    ConsoleGameBoyColor,
+                    ConsoleNES,
+                    ConsoleNintendo64,
+                    ConsoleGameCube,
+                    ConsolePokemonMini,
+                    ConsoleSNES,
+                    ConsoleSwitch,
+                    ConsoleVirtualBoy,
+                    ConsoleNewNintendo3DS
                 });
             Movies.SubCategories.AddRange(
                 new List<IndexerCategory>
--- a/src/NzbDrone.Core/Indexers/Definitions/GazelleGames.cs
+++ b/src/NzbDrone.Core/Indexers/Definitions/GazelleGames.cs
@@ -71,19 +71,19 @@ namespace NzbDrone.Core.Indexers.Definitions
             caps.Categories.AddCategoryMapping("Xbox 360", NewznabStandardCategory.ConsoleXBox360, "Xbox 360");
 
             // Nintendo
-            caps.Categories.AddCategoryMapping("Game Boy", NewznabStandardCategory.ConsoleOther, "Game Boy");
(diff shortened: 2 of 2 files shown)


mynameisbogdan:
@Taloth did you added new categories to the spec that aren't on https://torznab.github.io/spec-1.3-draft/external/newznab/api.html#predefined-categories ? 😸

src/NzbDrone.Core/Indexers/NewznabStandardCategory.cs
```diff
@@ -24,6 +24,17 @@ public static class NewznabStandardCategory
...
+        public static readonly IndexerCategory ConsoleGameBoyAdvance = new(1200, "Console/Game Boy Advance");
+        public static readonly IndexerCategory ConsoleGameBoyColor = new(1210, "Console/Game Boy Color");
+        public static readonly IndexerCategory ConsoleNES = new(1220, "Console/NES");
+        public static readonly IndexerCategory ConsoleNintendo64 = new(1230, "Console/Nintendo 64");
+        public static readonly IndexerCategory ConsoleGameCube = new(1240, "Console/Nintendo GameCube");
+        public static readonly IndexerCategory ConsolePokemonMini = new(1250, "Console/Pokemon Mini");
+        public static readonly IndexerCategory ConsoleSNES = new(1260, "Console/SNES");
+        public static readonly IndexerCategory ConsoleSwitch = new(1270, "Console/Switch");
+        public static readonly IndexerCategory ConsoleVirtualBoy = new(1280, "Console/Virtual Boy");
+        public static readonly IndexerCategory ConsoleNewNintendo3DS = new(1290, "Console/New Nintendo 3DS");
```
mynameisbogdan, line 37: From where did you pulled these ids? They aren't on https://torznab.github.io/spec-1.3-draft/external/newznab/api.html#predefined-categories

mynameisbogdan (changes requested):

src/NzbDrone.Core/Indexers/NewznabStandardCategory.cs
chickenhuntr: lol, ngl I didn't know they came from somewhere

Any other way to be able to do searches on the categories properly? 

Do they need to be added to torznab first I guess ?

chickenhuntr: Looking at that spec, it does not have WiiU, PS4, PS3 or XBox One - but prowlarr has these defined as selectable categories

mynameisbogdan: Well, how do you added them then? How other apps will know how to use those ids then? Prowlarr isn't the source of truth for them.

Taloth: The newsnab spec linked in the torznab spec, is a partial copy of the original newznab spec. (a copy of it seems to be here https://newznab.readthedocs.io/en/latest/misc/api.html)
But then an earlier version.

nZEDb's version of the newznab api.
https://github.com/nZEDb/nZEDb/blob/0.x/docs/newznab_api_specification.txt
You can see some more categories there.

Then you can check existing newznab sites to see if there is any consistency in new categories. Conflicts should be avoided.

You can already see that prowlarr xbox one (1140) doesn't match newznab's (1090).

1090 xbox one (newznab upstream, Usenetcrawler)
1140 xbox one (nZEDb upstream, Prowler, drunkenslug)

So I suspect Prowlarr copied nzedb list (which is somewhat sane coz nzedb was ahead at one point, but alas an inconsistency)

Anyway, I'm not sure if there should be a one-to-one mapping for each of them and polute the 'standard' list with something a specific site has. But that's for Powlarr devs to decide, not me. (I'm not from around here, I was Sonarr dev and created Torznab spec back in the day to standardize it for some torrent sites).
The spec basically says to have custom categories in the custom range 100000+, ultimately it doesn't matter much but i'd steer away from simply incremental numbers, as it _will_ eventually clash with nzedb's list and imo following some lead is sensible.

chickenhuntr: Yeah as I said, I didn't know they came from some where. I just incremented up by 10 as that's how it looked like it was done. 

Where did the ids come from for PS4 etc? They were added in this pr. https://github.com/Prowlarr/Prowlarr/pull/5/changes#diff-23aa29e04ca36409c79508d3513828312e4ff3f5a8413b5405dd359b87fab601

I just want better searching capabilities for GGn as right now if you search for something, it is riddled with 200+ more results not to do with the console you're searching for. From my testing, the changes in this PR resolved that.

Taloth: ps4 is listed as such in nzedb spec that I linked, so presumably it came from there.

chickenhuntr: HAve just seen Taloths message after my other one
> i'd steer away from simply incremental numbers

Would a sane approach be to start these category numbers that are not defined in the spec at 1500, and increment from there? It would stay within the 'console', and not go into the 2000 range which is for movies?

As far as I can tell for GGn, it is just sending the word 'Switch' (or whatever category mapping is defined.. Wii, Wii U, etc..) into the artistcheck array which is category on there. 

sorry, just seen your next message taloth, yeah I didn't see your original one when I was typing that other one :)

Taloth: No worries mate, I suspected the messages crossed a bit.

I honestly don't know enough about Prowlarr to say what should be done.

Officially, in newznab, each release exists in multiple categories (such as 1000, 1180). Meaning that a "New Nintendo 3DS" release should probably be in 1000, 1999, and 1000xx (eg 100020). (Dunno if that's how prowlarr works though coz it does aggregation). That way someone searching by default for 1000, or 1999, still gets those releases, or can filter just to 1000xxx

The one thing I'm a bit annoyed about is Switch not having an 'official' id yet. Given it's fairly mainstream.

Then it's about meaningfulness of having separate categories.
Virtual Boy, for example existed for 2 years. How big is the chance there are meaningful releases there that people _want_. So what's the argument of a separate category for it rather than Other.
"PC Games" is a single category and contains DOS games, Windows 95 games, Windows 10 games etc.
Might be true for others as well.

I'm just here to give some history and my 2 cents 😄 

@mynameisbogdan Ultimately you can do whatever. I dunno if apps other than Sonarr pull caps from prowlarr. And how it works.

chickenhuntr: Yes I always found it odd that switch, NDS, 3DS didn't have a category here. It was very hard to download via scripts or prowlarr UI. I ended up having to browse and it was not enjoyable :-) 

I agree on Virtual Boy, even the PokemonMini I would think. Things like DS, 3DS, GB, GBA, GBC, I can think of *many* titles personally - much more than Wii/WiiU who have their own category already.

If these cannot be added in, I will just integrate with GGn API in scripts myself I guess - but these extra categories would just give prowlarr yet another use

I am happy to do changes in the PR

mynameisbogdan: They do, but I'm not aware of any big ones focused on games. 

In any case, randomly adding them isn't something we should be doing and not so many at one time.

chickenhuntr: > They do, but I'm not aware of any big ones focused on games.
> 
> In any case, randomly adding them isn't something we should be doing and not so many at one time.

Does this not only affect GGN? it's the only indexer who gets the categories mapped?

IDK let me know if there's something I can do to get it approved, or just close it

mynameisbogdan: It does only affect GGn at the moment, but the ids aren't standardized or in use by other apps to match same exact id to same category names. 

While it's easy to merge them, it's a nightmare to change them afterwards if some apps follow these ids but we eventually need to fix them to something standardized. 

> IDK let me know if there's something I can do to get it approved, or just close it

Unless you can find a few indexers that follow these ids already, otherwise I don't think there is anything to be done, as the categories are actually generated. 🤷

chickenhuntr: Hello @mynameisbogdan, all good. 

I will just close it. 

I can see how it's not the correct approach. I will create a feature request, perhaps a better idea is to expose the custom ids (the ones taloth mentioned earlier) to the drop-down in the UI. They are actually auto generated as well, and from brief testing / a hacked together solution it seems like it'll work.

PR #2755, opened 2026-07-27
New: Show indexer specific categories in search

chickenhuntr:
Show custom categories when a single indexer is selected. Clear them if another indexer is selected.

#### Database Migration
NO

#### Description

Expose the custom categories to the UI search when **ONE** Indexer is selected. 
When two are selected, it will clear the custom categories.

I chose to make this only work for one indexer selected at a time because: 
1. It uses a shared category list on the search
2. This would then search Indexer A with Indexer B Custom categories
3. To prevent this would require backend changes instead of just front end

I think just being able to search one indexer in this case is perfectly fine, as if you're drilling down to a indexers categories you're probably only doing it for that specific indexer anyway. Standard categories are still selectale 

This is an alternative to : https://github.com/Prowlarr/Prowlarr/pull/2753 

Reason for something like this:

> Using Console/3DS for Pokemon X = 0 Results. It passes 3DS instead of 'Nintendo 3DS' into the query 
> So you would need to use Console/Other -> 162 Results, from wrong consoles
> Using the 'Nintendo 3DS' category, that could be selected from the UI (but cannot currently do, but can do in this branch)
> 65 Results, Only groups that reference "Pokemon X" are returned - Pokemon X & Y, and 3DS megapacks
>= much more usable

image
image

#### Todos
- [x] Tests

n/a

### DIFF
--- a/frontend/src/Components/Form/NewznabCategorySelectInputConnector.js
+++ b/frontend/src/Components/Form/NewznabCategorySelectInputConnector.js
@@ -8,7 +8,9 @@ function createMapStateToProps() {
   return createSelector(
     (state, { value }) => value,
     (state) => state.settings.indexerCategories,
-    (value, categories) => {
+    (state) => state.indexers.items,
+    (state, { indexerIds }) => indexerIds,
+    (value, categories, indexers, indexerIds) => {
       const values = [];
 
       categories.items.forEach((element) => {
@@ -29,6 +31,31 @@ function createMapStateToProps() {
           });
         }
       });
+
+      if (indexerIds && indexerIds.length === 1) {
+        const indexer = indexers.find((i) => i.id === indexerIds[0]);
+        const indexerCategories = (indexer?.capabilities?.categories ?? [])
+          .filter((category) => category.id >= 100000);
+
+        if (indexerCategories.length) {
+          const customCategoryGroup = {
+            key: `indexerCategories-${indexer.id}`,
+            value: indexer.name,
+            isDisabled: true
+          };
+
+          values.push(customCategoryGroup);
+
+          indexerCategories.forEach((category) => {
+            values.push({
+              key: category.id,
+              value: category.name,
+              hint: `(${category.id})`,
+              parentKey: customCategoryGroup.key
+            });
+          });
+        }
+      }
 
       return {
         value: value || [],
@@ -60,7 +87,7 @@ class IndexersSelectInputConnector extends Component {
 
 IndexersSelectInputConnector.propTypes = {
   name: PropTypes.string.isRequired,
-  indexerIds: PropTypes.number,
+  indexerIds: PropTypes.arrayOf(PropTypes.number),
   value: PropTypes.arrayOf(PropTypes.number).isRequired,
   values: PropTypes.arrayOf(PropTypes.object).isRequired,
   onChange: PropTypes.func.isRequired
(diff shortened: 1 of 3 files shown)


frontend/src/Search/SearchFooterConnector.js
```diff
@@ -80,6 +80,19 @@ class SearchFooterConnector extends Component {
...
+      );
+
+      this.props.setSearchDefault({
+        [name]: value,
+        searchCategories
+      });
+
+      return;
+    }
+
```
mynameisbogdan, line 95: What this for? Can't figure it out, since without it works too.

frontend/src/Components/Form/NewznabCategorySelectInputConnector.js
```diff
@@ -30,6 +33,32 @@ function createMapStateToProps() {
         }
       });
 
+      if (indexerIds && indexerIds.length === 1) {
+        const indexer = indexers.find((i) => i.id === indexerIds[0]);
+        const indexerCategories = (indexer?.capabilities?.categories ?? [])
+          .filter((category) => category.id >= 100000)
+          .sort(sortByProp('name'));
```
mynameisbogdan: While it nice to be have them sorted by name, see the indexer toloka where with many categories is nicer to preserve the order from the definition especially when nested. But yet again, not all are the same, some are a mess in line with the site.

frontend/src/Search/SearchFooterConnector.js
chickenhuntr: Hi B,
if you do the following without the snippet : 
1. you select an Indexer
2. select a custom category
3. un select the indexer (or select another one so there is multiple selected)

You will see that the categories actually stayed selected. This clears the custom category selections when changing indexers. here are two videos

Without:

https://github.com/user-attachments/assets/c770e851-2709-454e-83de-2821c04b4912

With :

https://github.com/user-attachments/assets/51f1b547-c24d-44e2-8f1d-f94081a34867

frontend/src/Components/Form/NewznabCategorySelectInputConnector.js
chickenhuntr: I don't have toloka or the other one you mentioned rutracker. is there any publics with this issue I can look at?

frontend/src/Search/SearchFooterConnector.js
mynameisbogdan: Ah, right!

frontend/src/Components/Form/NewznabCategorySelectInputConnector.js
mynameisbogdan: Add rutracker.org in prowlarr as disabled, then bulk edit and change just enable to true, apply. Then go search and rutracker should be available as a selectable indexer and enjoy the many categories it offers.

chickenhuntr: I've removed the sort, even the trackers I was testing with were nicer w/o it

RuTracker wasa bust, I couldnt understand any of the words

mynameisbogdan: rutracker was only to see how slow or fast is going to be the UI to render the dropdown with a larger amount of categories.

mynameisbogdan (approved):

augustuen (approved):

PR #2758, opened 2026-07-29
New: Added Sportarr as an application

Sportarr:
#### Database Migration

NO

#### Description

Adds Sportarr as a native application type. Sportarr is a sports PVR in the arr family, where leagues are the library items and events are the episodes, and this module gives it the same first class sync the other arr apps have instead of piggybacking on the Sonarr application type.

The implementation mirrors the Sonarr and Radarr modules: schema driven indexer building, full sync and add and remove only levels, category intersection, retained user fields on update, and ownership checks via the Prowlarr URL and API key. It talks Sportarr's application API, which is documented and version frozen in the Sportarr repo (docs/APPLICATION_API.md). The connection test checks the reported version and requires Sportarr 4.0.1023 or later with a clear message on older installs. I develop Sportarr, so the server side of this contract is maintained and versioned by me.

Tested against real builds end to end: add and test, version gate rejection on an older Sportarr, indexer sync with correct proxy URLs and categories, steady state with no redundant updates across repeated syncs, single update on real drift with user set fields retained, removal, and manually added Sportarr indexers left untouched. Full unit suite passes and the build is clean.

Opening this per the go ahead from the mods in the Discord prowlarr channel. Sportarr has been shipping integrations with the wider ecosystem recently, with native support merged in Maintainerr and Homepage, catalog listings in TrueNAS SCALE and HexOS, and an autobrr integration in review, and Prowlarr is the piece users ask about most.

A note on process, since it deserves saying plainly. I used an AI coding assistant to help draft this module and it is modeled closely on the existing Sonarr and Radarr implementations. I reviewed everything and verified the whole flow against live instances before opening this.

Sportarr in the Applications settings

#### Todos

- [ ] Tests (checked the existing application modules; none carry unit tests, so none are added here, happy to add some if you'd like them)
- [ ] Translation Keys (no new keys; the settings fields use the same labels the sibling application modules use)
- [ ] [Wiki Updates](https://wiki.servarr.com) (happy to write the Sportarr application page after merge)

None linked; requested and approved via Discord.


Sportarr:
The four failed integration jobs are all the same test, bad_ssl_should_pass_if_remote_validation_disabled in Common.Test, which live-checks expired.badssl.com. 

This PR only adds files under Core/Applications/Sportarr, and the actual integration suite passed on every runner. The badssl tests also pass locally from this branch, so it looks like the runners briefly had trouble reaching badssl.com. 

Could someone kick off a re-run when convenient? Thank you!

RetroDadson:
> The four failed integration jobs are all the same test, bad_ssl_should_pass_if_remote_validation_disabled in Common.Test, which live-checks expired.badssl.com.
> 
> This PR only adds files under Core/Applications/Sportarr, and the actual integration suite passed on every runner. The badssl tests also pass locally from this branch, so it looks like the runners briefly had trouble reaching badssl.com.
> 
> Could someone kick off a re-run when convenient? Thank you!

Kicked them off again, although I would like the Commit message changed to "New: Added Sportarr as an application" or similar so it's a bit more descriptive than just "Sportarr app"

Sportarr:
Thank you! Updated to "New: Added Sportarr as an application" as requested.

FrickDude:
Good Morning @RetroDadson - I see we're about a month from the last update. Obviously the review length of time would very widely, but out of curiosity, what is the approximate wait time to wait for a review to get started for something like this? Thank you.

PR #2781, opened 2026-08-15
Fixed: Stats page indexer/tag filters missing 'does not contain' option

### LINKED ISSUE #2617: Missing filter

### Current Behavior

Stats page custom filter missing other filter options. Only gives the option 'contains'.

### Expected Behavior

Dropdown has 'not contains' like the dropdown in custom filter on indexers page.

### Steps To Reproduce

http://prowlarr.lab/indexers/stats -- stats page
add custom filter
filter: Idexers, option dropdown here. 

### Environment

```markdown
- OS: debian13
- Prowlarr: 
- Docker Install: no 
- Using Reverse Proxy: no
- Browser: firefox dev

Version
    2.3.0.5236
.NET
    Yes (8.0.12)
Database
    Sqlite 3.50.4
Database Migration
    43
AppData Directory
    /var/lib/prowlarr
Startup Directory
    /opt/Prowlarr
Mode
    Console
```

Master

### Trace Logs?

n/a

chrisdebian:
Fixes #2617

## Root cause

The Stats page's `indexers` and `tags` filter fields (`frontend/src/Store/Actions/indexerStatsActions.js`) used `filterBuilderTypes.CONTAINS`, which only exposes a single `contains` operator (see `possibleFilterTypes` in `frontend/src/Helpers/Props/filterBuilderTypes.js`).

The main Indexers page (`frontend/src/Store/Actions/indexerIndexActions.js`) already uses `filterBuilderTypes.ARRAY` for the equivalent `tags`/`categories` fields, which exposes both `contains` and `does not contain`.

## Fix

Changed the Stats page's `indexers` and `tags` fields from `CONTAINS` to `ARRAY`, matching the existing pattern used on the Indexers page. Two-line change, purely frontend — all filtering for these fields happens client-side (`filterTypePredicates.js`), so no backend changes are needed.

## Testing

- `npx eslint` clean on the changed file
- No existing frontend test suite covers this file (repo has no `.test.js`/`.test.ts` files under `frontend/src`), so no test added, consistent with current coverage

### DIFF
--- a/frontend/src/Store/Actions/indexerStatsActions.js
+++ b/frontend/src/Store/Actions/indexerStatsActions.js
@@ -58,7 +58,7 @@ export const defaultState = {
     {
       name: 'indexers',
       label: () => translate('Indexers'),
-      type: filterBuilderTypes.CONTAINS,
+      type: filterBuilderTypes.ARRAY,
       valueType: filterBuilderValueTypes.INDEXER
     },
     {
@@ -70,7 +70,7 @@ export const defaultState = {
     {
       name: 'tags',
       label: () => translate('Tags'),
-      type: filterBuilderTypes.CONTAINS,
+      type: filterBuilderTypes.ARRAY,
       valueType: filterBuilderValueTypes.TAG
     }
   ],


mynameisbogdan:
Pretty sure the filtering happens server-side, Claude.

chrisdebian:
Good catch, thank you. I traced this further and you're right.

The Stats page's `fetchIndexerStats` sends `indexers`/`tags` as query params to the live `/indexerStats` endpoint rather than filtering a locally-fetched collection. `IndexerStatsController.GetAll` only ever intersects against the supplied IDs — there's no exclusion path in the backend at all. So this PR's frontend-only change would let a user select "does not contain" in the UI, but the request sent to the server would still be treated as an inclusion list, giving the opposite of the selected filter.

A correct fix needs backend changes (an actual exclude parameter and query logic, not just intersect), which is more than this PR set out to do. Closing this for now rather than pushing a bigger change on our first PR here — appreciate you flagging it before it merged.

Issue #2590, opened 2026-01-29
Add per indexer search mode override (force query search vs ID based search)


CalumOchkas:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Hi team,

I am running into an issue where Radarr sends ID based searches to Prowlarr for certain indexers, for example {imdb:tt318445xxx}.

The problem is that some indexers do not actually **consistently** return correct results for ID based searches. Because imdbid is advertised as supported, Radarr prefers ID based searches. As a consequence, Radarr/Sonarr manual search via Prowlarr can return no results for that indexer, while a plain text query directly on Prowlarr does.

An example indexer is TorrentDay, the native indexer struggles with IMDb-only search logic. It appears that while the tracker supports IMDb IDs in theory, a large volume of their uploads lack the proper tagging for an ID search to catch them.

This creates a 'false negative' where Prowlarr reports no results to Sonarr, even though the content is available.

### Describe the solution you'd like

Please add an [advanced] option to configure, per indexer, the search capabilities Prowlarr should present when receiving searches from apps like Radarr and Sonarr. This could be:

- an option to modify the list of search capabilities per indexer
- a checkbox to exclude certain capabilities

Seems like a similar solution has been configured here specifically for this indexer: https://github.com/Jackett/Jackett/commit/a1a98f98367f582d1e0fd6fe07eed11108d1ae9e

### Describe alternatives you've considered

None, alternative is to create custom definitions and manually set the search capabilities.

### Anything else?

N/A

augustuen:
The discussion from #2569 still applies. We prefer to fix the offending indexers rather than giving users a way to bypass it for all. This way one user's report can benefit the whole user base.

CalumOchkas:
But what if it's only a handful of their torrents which don't list the correct ID, and a number of Prowlarr users would prefer to keep searching TorrentDay by ID only? If the ID is not available via torrentday they would prefer to go elsewhere. Whereas for myself and some others, they prefer to have a fallback option available.

What's the aversion to doing this other than effort?

augustuen:
> But what if it's only a handful of their torrents which don't list the correct ID, and a number of Prowlarr users would prefer to keep searching TorrentDay by ID only? If the ID is not available via torrentday they would prefer to go elsewhere.

Those users would keep ID-based searches enabled for TorrentDay, like with the fix ilikeburning referenced in that other issue. 

[...]

Issue #2569, opened 2025-12-17
Add per indexer search mode override (force query search vs ID based search)


rodrigocabraln:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

<img width="477" height="298" alt="Image" src="https://github.com/user-attachments/assets/a975175e-88d7-4ffd-beef-7e8dfde24295" />

Hi team,

I am running into an issue where Radarr sends ID based searches to Prowlarr for certain indexers, for example {imdb:tt318445xxx}.

The problem is that some indexers do not actually return correct results for ID based searches. As a consequence, Radarr manual search via Prowlarr can return no results for that indexer, while a plain text query directly on prowlarr does

### Describe the solution you'd like

Please add an option to configure, per indexer, which search mode Prowlarr should use when receiving searches from apps like Radarr and Sonarr:

1 Force query based search
Always translate incoming searches to use the text query parameter q (for example q=xxxxxxxx) even if an ID is present.

2 Force ID based search
Always prefer imdbid, tmdbid, etc when available.

3 Auto (current behavior)
Use the current capability driven logic.

### Describe alternatives you've considered

None

### Anything else?

None

ilike2burnthing:
If indexers aren't working properly, then this should be reported and fixed, not worked around on a per-user basis.

What indexers are currently not working with ID search for you?

rodrigocabraln:
@ilike2burnthing 

This is not really an indexer problem, it is a small detail in the Torznab capabilities that affects search results.

Radarr reads the indexer caps first. In my case the caps include something like this:

```
<searching>
  <movie-search available="yes" supportedParams="q,imdbid" />
</searching>
```

Because imdbid is advertised as supported, Radarr prefers ID based searches. So when I run a manual search, Radarr sends only the IMDb ID to Prowlarr and does not perform a title query with q. For this specific indexer, that behavior is not ideal because I lose good results that only show up with a normal title search.


What I did to fix it

I put a simple Nginx reverse proxy in front of Prowlarr and rewrote the caps response. You need the ngx_http_subs_filter_module enabled.

Original response from Prowlarr:

```
curl -s "http://10.1.1.51:9696/14/api?t=caps&apikey=xxxxxxx" | sed -n '1,120p'

<movie-search available="yes" supportedParams="q,imdbid" />

```

The proxy replaces that so it becomes:

`<movie-search available="yes" supportedParams="q" />`


[...]

PR #2793, opened 2026-08-27
New: (IPTorrents) Option to disable IMDb ID search

LesIT1:
#### Database Migration
NO

#### Description
IPTorrents only matches an IMDb ID search when the uploader has filled in the IMDb field on the release, and plenty of uploads never get tagged. An ID search can therefore return a couple of tagged releases while missing better untagged ones, and since Radarr/Sonarr only fall back to a text search when the ID search comes back empty, those few tagged hits mask everything else. Real example from this week: a movie ID search returned 3 tagged TELESYNC rips while the site also had untagged WEB-DLs that never showed up.

This adds an opt-in "Disable IMDb ID Search" checkbox to the IPTorrents settings, the per-indexer approach suggested in #2569 and #2590 (and what Jackett did for the same problem in Jackett/Jackett@a1a98f98367f582d1e0fd6fe07eed11108d1ae9e). When enabled the indexer stops advertising `imdbid` in its TV/movie caps, so apps send title searches instead, and the request generator also ignores any IMDb ID it still receives. Default is off, which leaves current behaviour exactly as it is.

Implementation note: the capabilities read the setting through `Definition?.Settings` with a null guard, because template instances get their caps queried before a definition is assigned (`ProviderFactory.GetDefaultDefinitions`).

The label key is the generic `IndexerSettingsDisableImdbSearch` so other indexers with the same tagging problem (TorrentDay per #2590) can reuse it later.

Settings checkbox renders from the field definition. Happy to add a screenshot if wanted.

#### Todos
- [x] Tests

* Related discussion: #2569, #2590 (closed; per-indexer fix was the suggested path)


mynameisbogdan:
> This adds an opt-in "Disable IMDb ID Search" checkbox to the IPTorrents settings, the per-indexer approach suggested in #2569 and #2590 (and what Jackett did for the same problem in [Jackett/Jackett@a1a98f9](https://github.com/Jackett/Jackett/commit/a1a98f98367f582d1e0fd6fe07eed11108d1ae9e)). When enabled the indexer stops advertising `imdbid` in its TV/movie caps, so apps send title searches instead, and the request generator also ignores any IMDb ID it still receives. Default is off, which leaves current behaviour exactly as it is.

It's not exactly the same apples to apples here, since the example you provided still relies on tmdbid.

src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
```diff
@@ -221,7 +233,7 @@ private IEnumerable<IndexerRequest> GetPagedRequests(string term, SearchCriteria
                 }
             }
 
-            if (imdbId.IsNotNullOrWhiteSpace())
+            if (imdbId.IsNotNullOrWhiteSpace() && !Settings.DisableImdbSearch)
```
mynameisbogdan, line 236: Did you tested these changes? 

This change would basically act like an keyword-less search, and since there are 100 items per page this will paginate up to 10 pages, so that's why I'm asking, did you tested these changes?

src/NzbDrone.Core/Indexers/Definitions/IPTorrents.cs
```diff
@@ -82,15 +82,21 @@ protected override IList<ReleaseInfo> CleanupReleases(IEnumerable<ReleaseInfo> r
 
         private IndexerCapabilities SetCapabilities()
         {
+            // The site only matches an IMDb ID if the uploader has tagged the release with it,
+            // so the ID search caps are removed when the user opts out of IMDb ID searches.
+            // Settings are accessed via the definition because the capabilities of template
+            // instances are queried before a definition is assigned.
+            var disableImdbSearch = Definition?.Settings is IPTorrentsSettings settings && settings.DisableImdbSearch;
```
mynameisbogdan, line 89: Sadly this feels like a hack at best. At this point just remove the imdb search completely and be done with it.

PR #2814, opened 2026-09-07
Add basic api search cache

### LINKED ISSUE #2272: Caching

Sometimes I like to manually review the choices before letting the system pick.  Even done within a minute of each other, this requires the indexers to be queried again. 

Can we allow a cache?  The user can specify how long they want the cache to remain, or whether they want it at all.  In searching the user could be shown the cached results and also be given an option to perform a new query.

Presently, if I want to review the sources, I do a manually search, wait a while for results, and then view it.

If I am going to do the automatic search, I’ll still need to wait for the system to get the results again from the indexers to download it.

### Anything else?

I’m interested to see if others think this could be useful. If I have a peculiar way of using prowlarr with my ARRs, it might not be helpful for many, so feel free to close this.

Namyts:
#### Database Migration
NO

#### Description

When using Prowlarr it **is a very common pattern** to:
- search (wait a while, make indexer requests)
- select 1
- the 1 selected fails for various reasons
- you want to search again (wait a while, and make the exact same indexer requests)

It is a complete waste to pull 100s of results each time, instead of easily being able to select the next one in the previous list.

This could also be implemented on the searchees side, but makes sense to be implemented here first since it also is an easy dedupe within the cache ttl.

If you are worried about high memory usage - let's work on adding a simple quantity cap before results get expunged

image

#### Todos
- [x] Tests

* Fixes https://github.com/Prowlarr/Prowlarr/issues/2272


mynameisbogdan:
Serious question, why everyone is implementing their cache solution like it's the simplest thing to do?

I suspect your solution, while supports only in-memory caching, it's prone to cache stampede.

I have some ideas related to caching, but sadly those need to wait for .NET 10 as HybridCache it's available for .NET 9 and up.

Namyts:
> like it's the simplest thing

I agree that caching can have additional complexities, but I also prefer to _not let perfect be the enemy of good_,
and I think that this change can positively benefit some users **today**... (it is also optional)

> only in-memory caching

Should be easy enough to improve the implementation later, or even add a dropdown to configure external caching (you mentioned redis in the Issue).

> cache stampede

Just added 1 more commit to reset the TTL on fetched items which should lessen that effect.

mynameisbogdan:
> I agree that caching can have additional complexities, but I also prefer to _not let perfect be the enemy of good_,
> and I think that this change can positively benefit some users **today**... (it is also optional)

When it's a single user, sure, let's not let perfect be the enemy of good, whatever this means, but when it affects multiple users you need a proper solution, not a half baked one, otherwise you're going to deal with TTL issues and incompatibilities between JSON resources.

> Just added 1 more commit to reset the TTL on fetched items which should lessen that effect.

I don't think that helps, the issue with cache stampede when an endpoint gets multiple hits in the same time and all try to generate the cached value before saving. Cache stampede protection is to allow only the first one to generate the value, the others will wait.

https://en.wikipedia.org/wiki/Cache_stampede

Namyts:
> Cache stampede

Added a waiting layer for identical simultaneous requests which should help with the stampede issue. However it is worth mentioning that without this cache there is currently no protection from a stampede anyway?

> TTL issues

Keeping this feature optional allows Prowlarr admins to decide...

>  sure, let's not let perfect be the enemy of good, whatever this means

I am just gently suggesting that I'd obviously prefer a solution today that acceptably resolves an issue I (and others?) am having, vs blocking until a solution with no imperfections arrives... 

I appreciate your work & quick replies! Are there any other improvements you want me to address for this initial implementation?

mynameisbogdan:
> is currently no protection from a stampede anyway?

Yea, there's rate limiting that prevents doing the same thing in the same time, but they'll still be executed after 2+ seconds. But we're talking cache stampede problems here, not how in general the APIs works. You're not fixing an general "stampede" issue, you're fixing a well known caching stampede issue.

> Are there any other improvements you want me to address for this initial implementation?

I have none, as I'm against adding a custom caching implementation. I've used HybridCaching for some time and quite happy with it how it natively supports L1 ad L2 caches, so you can add any external distributed caching storage you would like, fail-safe, tagging, etc. 

I have no issues pushing .NET 10 bumps to Servarr apps, but they need to fix their dotnet-freebsd build for .NET 10 and I'm not going to do that since I'm already building those on my own for Sonarr v5 and I'm not that familiar with azure pipelines, plus there's a very big delay in their bumps while mine are out usually in hours since I use renovate to manage the updates.

But I'm not a maintainer to decide, just a contributor like you.

Namyts:
> I have none, as I'm against adding a custom caching implementation

A shame. I have taken a look at the HybridCache library, and this PR could pretty easily pretty trivially switch to using it in the future when it becomes accessible.

> everyone is implementing their cache solution

In the meantime, I suspect more people will continue to submit similar PRs, if it is a desirable feature.

As for me: [I will just use a fork with these commits](https://github.com/Namyts/Prowlarr/pkgs/container/prowlarr) until Prowlarr supports search caching "properly"

Issue #27, opened 2021-03-18
Feature Parity with Hydra


rg9400:
- [ ] Stats:
  - [ ] 1. Uniqueness Stat
  - [ ] 2. Successful Downloads by indexer
  - [ ] 3. Recompute stats by using a multi-select category filter up top. So it can be possible to see stats for Books or Music + Audiobooks
- [ ] Search History:
  - [ ] 1. Show ID or RSS queries
  - [ ] 2. Show categories and source in table
- [ ] Download History:
  - [ ] 1. Show success/fail by download
  - [ ] 2. Allow regrab of specific download
- [ ] Downloader:
  - [ ] 1. Give option to configure Downloader, allow sending results to this downloader
    - [ ] 	* Create option to upload NZB to downloader or to send link for manual downloads
    - [ ] 	* Need to make sure internal links can be used when sending to downloader, even if Prowlarr is being used behind RP
  - [ ] 2. Give option to configure blackhole for torrents
  - [ ] 3. Handle magnet links for Torrents
  - [ ] 4. Update status of downloads from Downloader
  - [ ] 5. Potentially show download queue? Hydra does this, but seems not super important
  - [ ] 6. Give default category-wise options for manual downloads, with option to override
- [ ] Categories:
  - [ ] 1. Create custom categories
  - [ ] 2. Override search types
  - [ ] 3. Allow restrictions to be applied for API or manual searches or both
  - [ ] 4. Category-specific restrictions such as required words, regex, forbidden words, forbidden regex, size restrictions
  - [ ] 5. Ignore results
- [ ] Statuses:
  - [ ] 1. Show indexer API hits and downloads compared to limits
  - [ ] 2. Show current state, last error, and if disabled, when it was disabled and until when
- [ ] Searching:
  - [ ] 1. Timeout
  - [ ] 2. User Agent
  - [ ] 3. Map external user agents
  - [ ] 4. Ignore load limiting/errors
  - [ ] 5. Handle Torznab categories or ignore them
  - [ ] 6. Map indexer specific categories to internal categories
  - [ ] 7. Convert Media IDs for indexers (TMDB to IMDB for example)
  - [ ] 8. Generate queries when indexer does not support IDs
  - [ ] 9. Fallback to generated queries if no results found
  - [ ] 10. Global restrictions (word filters, forbidden posters, max results age, min seeders, ignore passworded releases)
  - [ ] 11. Option to remove trailing garbage (obfuscated, scrambled, postbot)
[...]

PR #2820, opened 2026-09-15
New: Configurable User-Agent for indexers

csmarshall:
#### Database Migration
YES - 045

#### Description
Adds a configurable User-Agent for indexers, in two layers, with precedence *per-indexer → global → built-in default*:

- a global User-Agent under **Settings → Indexers → Options**, applied to all indexer requests
- a per-indexer override on the indexer's edit form (advanced); blank inherits the global

Leaving both blank preserves current behaviour exactly.

This is the **Searching → 2. User Agent** line item from #27, claimed in [this comment](https://github.com/Prowlarr/Prowlarr/issues/27#issuecomment-5684042006). Opening the PR early per the contributing guide so there's something concrete to react to — I asked three design questions on that issue and they're repeated below, since any of them could change this.

**Implementation notes**

- Applied in `IndexerHttpClient.ExecuteProxied/ExecuteProxiedAsync`, the single path every indexer request goes through, so it covers search, login, captcha, download and capabilities in one place.
- Stored on `IndexerDefinition` next to `Priority`/`Redirect`/`DownloadClientId` rather than in the per-indexer settings contracts, so it applies to every indexer implementation without renumbering any `FieldDefinition` orders.
- Applied *before* the indexer proxies run, so a configured value wins over the per-host User-Agent FlareSolverr caches (it only injects that when the request has none). FlareSolverr still overrides on requests it actually solves.
- Validated on save with the same parser `ManagedHttpDispatcher` uses. Without that, an unparseable value saved cleanly and then failed every later request as "Unable to connect to indexer", with the real `FormatException` only visible in the log.
- The per-indexer field shows the effective default as its placeholder and fills it on focus so it can be edited rather than retyped, but clears it again on blur if left untouched — otherwise a stray click would pin that indexer to the current version string forever.

**Three decisions I'd like a call on**

1. **Ordering vs FlareSolverr** — a configured UA currently beats the solver's cached UA on the first attempt. Easy to flip if you'd rather the solver always win.
2. **Overriding a definition's UA** — a user-set value takes precedence over a `User-Agent` in a Cardigann definition's `headers:`. Deliberate (it's the escape hatch when a definition's UA goes stale), but it does mean someone can break rustorka/mojblink/etc. for themselves.
3. **Freeform vs presets** — freeform only, consistent with the existing IPTorrents and PixelHD fields. I'd argue against a dropdown of "common browser" UAs since they go stale every few weeks.

Happy to split this into two PRs (per-indexer first, global second) if that suits the one-feature-per-PR rule better.

shot-global
shot-global.png — Settings → Indexers. The global User-Agent, in an Options fieldset below Indexer Proxies to match Sonarr's Settings → Indexers layout. The placeholder shows the current (dynamically generated default `Prowlarr/10.0.0.20885 (debian 12)`.

shot-per-indexer
shot-per-indexer.png — the Edit Indexer modal with the per-indexer field directly under Indexer Priority, orange label marking it advanced, and the help text spelling out the precedence including the FlareSolverr caveat.

#### Todos
- [x] Tests

**Testing done**

- `IndexerHttpClientFixture` (9) covers the precedence chain and that the UA is applied before the proxies see the request.
- `045_indexer_user_agentFixture` (3) covers the migration.
- `IndexerResourceFixture` (5) covers the resource mapping. This needed a `ProjectReference` from `Prowlarr.Api.V1.Test` to `Prowlarr.Api.V1` — that project is named for Api.V1 but did not reference it, so the API layer had no unit coverage. Happy to drop that if it's unwanted.
- Also ran it in a container against a local Torznab endpoint that logs the `User-Agent` it receives, and confirmed each step of the precedence chain on the wire, plus that the migration applies to a fresh database.

Every test above was checked against a deliberately broken version as well as a working one, so they're known to be capable of failing.

* Part of #27

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mynameisbogdan:
Not a good idea considering indexers might keep track of the user-agents to detect duplicated accounts.

csmarshall:
Worth noting the capability already exists, just unevenly. IPTorrents and PixelHD both ship a user-settable "Cookie User-Agent" field today, and the definitions repo has merged UA overrides for rustorka (Prowlarr/Indexers#740), mojblink (Prowlarr/Indexers#769), soundpark (Prowlarr/Indexers#131) and gay-torrents (Prowlarr/Indexers#970). This generalises that rather than introducing it.

Those existing fields serve the opposite of evasion: the tracker wants the UA to match the browser session the cookie came from — `info_useragent` tells users to copy both from the same devtools panel. A UA that disagrees with its cookie is itself the anomaly.

Blank is the default and is byte-for-byte current behaviour, on an advanced field.

mynameisbogdan:
No, those are all torrent sites and it doesn't generalises anything when there's a difference between UA override needed to bypass cloudflare challenge/logic captcha versus this. Usenet indexers behave differently.

